PR #12, opened 2020-03-27
Updated README.md with running function on serverless platforms

meteatamel:


README.md
```diff
@@ -86,3 +86,14 @@ Cloud Run and Cloud Run on GKE both implement the [Knative Serving
 API](https://www.knative.dev/docs/). The Functions Framework is designed to be
 compatible with Knative environments. Just build and deploy your container to a
 Knative environment.
+
+## Configure the Functions Framework
```
jskeet: This information now occurs twice in the README - is it worth just replacing the existing occurrence with this section, perhaps?

meteatamel: Yes, I think it's nice to have a separate section for all configuration

jskeet: I'm fine with having it as a separate section - I just don't think it *also* needs to be in the "Run the function" section.

meteatamel: How about we merge "Run the function" section with the previous section and in the previous section, just show how to run it without getting into config details? This is also consistent with python and nodejs frameworks

jskeet: Yup, that's fine. I'm happy for you to do that in this PR, if you're happy to do it.

jskeet:
Merging, thanks!

PR #19, opened 2020-04-01
Simplify log output

jskeet:
- Remove the warning about overwriting a default value
- Use a one-line-per-message console writer, either simply text, or
  JSON if we detect we're on k-native via the K_SERVICE environment
  variable

Admittedly this loses the "prettiness" of coloured output etc - but it's a lot more appropriate for microservices in general. (And the JSON output looks great in Stackdriver. Once we've got an alpha02 nuget package, I'll deploy to Cloud Run to test that. It's tricky to do before then.)

### DIFF
--- a/Google.Cloud.Functions.Invoker/EntryPoint.cs
+++ b/Google.Cloud.Functions.Invoker/EntryPoint.cs
@@ -12,15 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Google.Cloud.Functions.Framework;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 using System;
-using System.Globalization;
-using System.Linq;
-using System.Net;
 using System.Reflection;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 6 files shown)


Google.Cloud.Functions.Invoker/EntryPoint.cs
```diff
@@ -54,15 +50,24 @@ public static class EntryPoint
...
             // TODO: Catch exceptions and return 1, or just let the exception propagate? It probably
             // doesn't matter much. Potentially catch exceptions during configuration, but let any
             // during web server execution propagate.
             var environment = FunctionEnvironment.Create(functionAssembly, args, ConfigurationVariableProvider.System);
-            var builder = Host.CreateDefaultBuilder(args)
+            IHostBuilder builder = Host.CreateDefaultBuilder(args)
                 .ConfigureWebHostDefaults(webBuilder => webBuilder
+                    .ConfigureLogging(logging =>
+                    {
+                        logging.ClearProviders(); logging.AddProvider(environment.LoggerProvider);
```
amanda-tarafa, line 66: nit: different line for adding the provider?

Google.Cloud.Functions.Invoker/EntryPoint.cs
```diff
@@ -54,15 +50,24 @@ public static class EntryPoint
...
+                    })
                     .ConfigureKestrel(serverOptions => serverOptions.Listen(environment.Address, environment.Port))
                     .Configure(app => app.Run(environment.RequestHandler))
-                );
+                ); ;
```
amanda-tarafa, line 70: nit: double ;

Google.Cloud.Functions.Invoker/EntryPoint.cs
jskeet: Doh - absolutely. Thanks!

Google.Cloud.Functions.Invoker/EntryPoint.cs
jskeet: Nicely spottted! Thanks :)

PR #21, opened 2020-04-01
If there's only one function in the calling assembly, use that by default

### LINKED ISSUE #20: Consider using a default function
If the entry point assembly only contains a single type implementing any of the functions interfaces, maybe we should use that by default.

This would be similar to how a console app looks for a single Main method entry point by default. Yes, you can override it, but you don't need to in the common case.

jskeet:
Fixes #20 and improves usability quite a bit :)

### DIFF
--- a/Google.Cloud.Functions.Invoker.Tests/FunctionEnvironmentTest.cs
+++ b/Google.Cloud.Functions.Invoker.Tests/FunctionEnvironmentTest.cs
@@ -94,6 +94,8 @@ namespace Google.Cloud.Functions.Invoker.Tests
             await AssertHttpFunctionType<DefaultFunction>(environment);
         }
 
+        // The test will pass because CreateEnvironment throws: the environment is "bad"
+        // because there are multiple function types in this assembly.
         [Fact]
         public void TargetFunction_Unspecified() =>
             AssertBadEnvironment(Empty, Empty);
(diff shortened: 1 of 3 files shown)


Google.Cloud.Functions.Invoker.Tests/FunctionEnvironmentTest.cs
```diff
@@ -94,6 +94,7 @@ public async Task TargetFunction_CommandLineOverridesEnvironmentVariable()
             await AssertHttpFunctionType<DefaultFunction>(environment);
         }
 
+        // This will throw (correctly) because there are multiple function types in this assembly.
```
amanda-tarafa, line 97: nit: Quickly reading this I understood that the test was throwing. Maybe change "This will throw (correctly)" by "This is a bad environment ..."

Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
```diff
@@ -61,6 +61,33 @@ internal sealed class FunctionEnvironment
         internal static FunctionEnvironment Create(Assembly functionAssembly, string[] commandLine, ConfigurationVariableProvider variableProvider) =>
             new Builder(functionAssembly, commandLine, variableProvider).Build();
 
+        /// <summary>
+        /// Attempts to find a single valid non-abstract function class within the given set of types.
+        /// (Note that "non-abstract class" is a pretty low bar; we could enforce non-generic etc, but we'll
```
amanda-tarafa, line 66: I agree this can wait but the error message they'll get might be confusing if the target type we find is generic for instance. It will fail when attempting to create an instance and some users might not even understand why an instance of that class is being created?

Google.Cloud.Functions.Invoker.Tests/FunctionEnvironmentTest.cs
jskeet: Yup. The comment was even worse the first time :)

Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
jskeet: Hmm. The error message isn't ideal, indeed:

> Unhandled exception. System.ArgumentException: Cannot create an instance of Google.Cloud.Functions.SampleHttpFunction.Function`1[T] because Type.ContainsGenericParameters is true.

But let's wait for a bit - I suspect we may want to change the instantiation model quite a bit in order to benefit from dependency injection, too... all of that could change things.

PR #22, opened 2020-04-02
Final steps for releasing 1.0.0-alpha02

jskeet:
(I could break this up into separate PRs if you'd prefer, but each commit is fairly straightforward.)

I've tested this with a local build of the packages, including deploying a Cloud Event Function to Cloud Run.


Google.Cloud.Functions.Framework.Tests/CloudEventAdapterTest.cs
```diff
@@ -0,0 +1,76 @@
...
+using CloudNative.CloudEvents;
+using Microsoft.AspNetCore.Http;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Functions.Framework.Tests
+{
+    /// <summary>
+    /// Sample no-op test just to get the infrastructure correct.
```
amanda-tarafa, line 24: nit: I think this comment is now obsolote.

jskeet: Whoops, that's very true. Will fix, thanks!

PR #25, opened 2020-04-03
Reorganize folder structure

jskeet:
(This will be one commit once #24 is merged.)


jskeet:
Note that github isn't very good at showing renames as far as I can tell - but almost everything here *is* just renamed.

jskeet:
Will rebase now #24 is merged.

PR #27, opened 2020-04-03
Minor build simplifications

jskeet:
(No user impact at all.)


jskeet:
(We might want to look at log redirection for test purposes at some point, but that's a different matter.)

PR #35, opened 2020-04-08
Add F# support in templates and invoker

### LINKED ISSUE #31: Feature: F# templates
It should at least be easy to create an F# template that still implements the interface defined in the Functions.Framework package. How idiomatic it will be is up for grabs. But let's try.

jskeet:
Fixes #31.

(First 5 commits will go away when #33 is merged.)


src/Google.Cloud.Functions.Templates/templates/http-function-fs/Function.fs
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace MyFunction
+
+open Microsoft.AspNetCore.Http
```
amanda-tarafa, line 17: Nit: alphabetically (I suppose)

jskeet: Done.

PR #49, opened 2020-04-23
Initial support for legacy events

jskeet:
PR notes:

- The first three commits are just fixes to minor problems I noticed while building the rest
- Still to do:
  - Potentially other event types
  - Documentation
  - F# and VB templates
  - A more interesting demo (working on this now...)

cc @grant @juliehockett

### DIFF
--- a/src/Google.Cloud.Functions.Framework/LegacyEvents/Context.cs
+++ b/src/Google.Cloud.Functions.Framework/LegacyEvents/Context.cs
@@ -1,0 +1,49 @@
+﻿// Copyright 2020, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Google.Cloud.Functions.Framework.LegacyEvents
+{
+    /// <summary>
+    /// The context of an event.
+    /// </summary>
+    public sealed class Context
+    {
+        /// <summary>
+        /// A unique ID for the event.
+        /// </summary>
+        [JsonPropertyName("eventId")]
+        public string? Id { get; set; }
+
+        /// <summary>
+        /// The date/time this event was created.
+        /// </summary>
+        [JsonPropertyName("timestamp")]
+        public DateTimeOffset? Timestamp { get; set; }
+
+        /// <summary>
+        /// The type of the event. For example: "google.pubsub.topic.publish".
+        /// </summary>
+        [JsonPropertyName("eventType")]
+        public string? Type { get; set; }
+
+        /// <summary>
+        /// The resource associated with the event.
+        /// </summary>
+        [JsonPropertyName("resource")]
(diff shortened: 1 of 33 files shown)


src/Google.Cloud.Functions.Framework/LegacyEvents/Context.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Google.Cloud.Functions.Framework.LegacyEvents
+{
+    /// <summary>
+    /// The context of an event.
+    /// </summary>
+    public class Context
```
amanda-tarafa, line 23: nit: This name is misleading? Maybe EventDescripto?

src/Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
```diff
@@ -149,8 +151,31 @@ internal static Type FindDefaultFunctionType(params Type[] types)
...
+        /// <summary>
+        /// Checks whether a given type implements a generic interface (assumed to have a single type parameter),
+        /// and returns the type argument if so. For example, if we have "class MyFunction : ILegacyEventFunction{PubSubMessage}"
+        /// then a call to GetGenericInterfaceImplementationTypeArgument(typeof(MyFunction), typeof(ILegacyEventFunction&lt;&gt;))
+        /// will return typeof(PubSubMessage).
+        /// </summary>
+        /// <param name="target">The target type to check.</param>
+        /// <param name="genericInterface">The generic interface</param>
+        /// <returns></returns>
+        private static Type? GetGenericInterfaceImplementationTypeArgument(Type target, Type genericInterface)
```
amanda-tarafa, line 169: If `target` were to implement the `genericInterface` twice with different type arguments, we would rely on `GetInterfaces()` being deterministic in its order for always returning the same type argument, which I think we shouldn't. Maybe we can order by type argument name and always return the first one.
For instance:
`class MultiLegacyEventFunction : ILegacyEventFucntion<StorageObject>, ILegacyEventFunction<PubSubMessage> {...}`
We should always return the same one of `StorageObject` or `PubSubMessage`.
Very very ideally, `LegacyEventAdapter` could handle two (many) payload types, but I do think that might be a lot of work for the very few times it will happen.
But being deterministic in what we return is easy enough to implement and it's really nice to have.

src/Google.Cloud.Functions.Framework/LegacyEvents/FirestoreEvent.cs
```diff
@@ -0,0 +1,267 @@
...
+            return map.Values!.Select(value => value.Value).ToList();
+        }
+
+        public override void Write(Utf8JsonWriter writer, IList<object?> value, JsonSerializerOptions options) =>
+            throw new NotImplementedException("FirestoreDocument serialization is not currently supported");
+    }
+
+    internal sealed class FirestoreValue
+    {
+        private Case _case;
```
amanda-tarafa, line 150: Totally fine with leaving this in, the more explicit the better.

src/Google.Cloud.Functions.Framework/LegacyEvents/Context.cs
jskeet: While I generally agree, this is the term that's generally used in documentation etc.
(The class should be sealed though - will fix that!)

src/Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
jskeet: I think we should fail in that case. We wouldn't know which adapter to route it to, without extra information. Will implement and add a test.
(I don't think the error message needs to be super-clear here, as it's extremely unlikely to happen. I'm happy to field issues on that front.)

PR #54, opened 2020-04-24
Provide a cancellation token for event functions

### LINKED ISSUE #50: Possibly provide a CancellationToken to event functions
We have a cancellation token from the "request aborted" part of the HttpContext - it would be good to pass that in.

jskeet:
Fixes #50.

(I'm happy to consider alternative options if you can think of them.)


jskeet:
> I wonder also if at some point you'd like to pass in more HttpContext stuff?

Hmm, not sure. Definitely not the request or response - we should be providing all of that. I wonder if there's anything else though. We can see over time. I guess that could probably be injected via DI too...

PR #55, opened 2020-04-24
Rename the TimeZoneConverter function to Function

### LINKED ISSUE #45: Rename TimeZoneConverter to Function
Currently the type has the same name as the namespace, which is a bad idea.

jskeet:
This is partly to avoid the "type name collision with namespace
name" aspect, and partly because I think I'm landing on Function
being a conventional entry point name in the same way that Program
is for a console application.

Fixes #45.


jskeet:
(Last PR before Monday, I promise :)

PR #58, opened 2020-04-27
PROTOTYPE ONLY: Sample of how "replying to an event" might be handled.

jskeet:
A different function might add a dependency on ICloudEventPublisher and publish any number of events during function execution.

(This is not intended to be merged - it's for the sake of discussion.)

This doesn't have nearly as much commentary as I'd like to include, but I've run out of time for the day. Hopefully it makes some kind of sense!

cc @meteatamel

### DIFF
--- a/examples/Google.Cloud.Functions.Examples.EventFunctionWithReply/Function.cs
+++ b/examples/Google.Cloud.Functions.Examples.EventFunctionWithReply/Function.cs
@@ -1,0 +1,50 @@
+﻿// Copyright 2020, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


meteatamel:
I like the simplicity of the API!

grant:
Event replying sounds good.
I really don't understand C# or have a history with library's structure so I can't provide a review. Design docs would be better for feedback.

meteatamel:
Note that I recently wrote 2 samples with Knative Eventing ([sample1](https://github.com/meteatamel/knative-tutorial/blob/master/docs/image-processing-pipeline.md), [sample2](https://github.com/meteatamel/knative-tutorial/blob/master/docs/bigquery-processing-pipeline.md)) and both used event replies to pass along messages. It would have been nice to use Functions Framework for them. This is to say that event replies are quite useful in any kind of event-driven pipelines.

grant:
> Note: Unassigning inactive GitHub issues. Re-assign when ready.

This general idea sounds good. We should create a GitHub issue/PR in the FF contract [0] for an optional feature first.

[0] https://github.com/GoogleCloudPlatform/functions-framework

jskeet:
Closing as I'm not expecting us to do this for beta, and it's now rather ancient. (Recreating is probably simpler than rebasing.)

PR #61, opened 2020-04-29
DRAFT: One possible approach for simpler errors.

### LINKED ISSUE #37: Work out what to do about communicating errors from functions to callers
In particular, what do we do beyond status codes? Where do we put messages?
What happens if you've written something to the body already?

jskeet:
Would fix #37 if we decided to do this. We want to review this as a possible cross-platform approach though.

### DIFF
--- a/src/Google.Cloud.Functions.Framework/FunctionException.cs
+++ b/src/Google.Cloud.Functions.Framework/FunctionException.cs
@@ -1,0 +1,65 @@
+﻿// Copyright 2020, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Net;
+
+namespace Google.Cloud.Functions.Framework
+{
+    // More decisions and considerations:
+    // - Another message to be set on the response? (As content? Something else?)
+    // - Prohibit success status codes?
+    // - What happens if we've already started writing a streaming response? (Is that feasible?)
+    // - What's the best way of controlling the logging?
+    // - Use an int instead of HttpStatusCode, as per HttpContext.Response.StatusCode?
+    
+    /// <summary>
+    /// Exception to provide a convenient way of aborting a function with a specific HTTP status code.
+    /// When a function is executed via Google.Cloud.Functions.Invoker package, this exception is caught
+    /// at the top level, the message is logged, and the response status code is set accordingly.
+    /// This exception should generally not be caught within function code.
+    /// </summary>
+    public class FunctionException : Exception
+    {
+        /// <summary>
+        /// The status code to set in the response.
+        /// </summary>
+        public HttpStatusCode StatusCode { get; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


jskeet:
Note: we could potentially add an adapter to the Google.Cloud.Functions.Framework type that basically had the code that's in FunctionEnvironment.Execute. That would make it easier to get a consistent experience when *not* using the Invoker package.
Let's worry about those implementation details once we've decided on the user experience in general.

src/Google.Cloud.Functions.Framework/FunctionException.cs
```diff
@@ -0,0 +1,65 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Net;
+
+namespace Google.Cloud.Functions.Framework
+{
+    // More decisions and considerations:
+    // - Another message to be set on the response? (As content? Something else?)
```
amanda-tarafa, line 21: I don't like this. It might be used to write random content, specially if success status codes are allowed.

src/Google.Cloud.Functions.Framework/FunctionException.cs
```diff
@@ -0,0 +1,65 @@
...
+    // - Prohibit success status codes?
```
amanda-tarafa, line 22: I'd say yes, although maybe provide another way to set different success status codes? But definitely wouldn't like an exception controlling different success statuses.

src/Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
```diff
@@ -126,12 +126,25 @@ internal void ConfigureServices(IServiceCollection services)
...
         internal async Task Execute(HttpContext context)
         {
-            var function = context.RequestServices.GetRequiredService<IHttpFunction>();
-            await function.HandleAsync(context);
+            try
+            {
+                var function = context.RequestServices.GetRequiredService<IHttpFunction>();
+                await function.HandleAsync(context);
+            }
+            catch (FunctionException ex)
```
amanda-tarafa, line 138: Maybe catch everything else as well to log a rethrow?

src/Google.Cloud.Functions.Framework/FunctionException.cs
jskeet: Well users can already write whatever content they want... it just gives an easy way to give callers a hint as to what's wrong (e.g. which required parameter is missing)

src/Google.Cloud.Functions.Framework/FunctionException.cs
jskeet: It would certainly be an anti-pattern to do so. Whether that makes it worth explicitly prohibiting or not, I don't know... but I'm happy enough to do it.

src/Google.Cloud.Functions.Invoker/FunctionEnvironment.cs
jskeet: No need for that - any exceptions that make it to the very top of the pipeline are logged and the response becomes a 500.

src/Google.Cloud.Functions.Framework/FunctionException.cs
amanda-tarafa: Yes, that's true, I guess what I really don't like is the combination of this and being able to set success status code from the exception. I can imagine it being used to control flow and that makes me icky.
I like a lot what MVC has with the IActionResult and all sorts of derived types, even types for bad status code like UnauthorizedResult, NotFoundResult, etc. It gave the opportunity to cleanly control for those cases I think.

grant:
Not sure if I am the best person to review these changes. Going to unassign, but feel free to ping me on docs or other non .NET specific changes.

jskeet:
Closing as I'm not expecting us to do this for beta, and it's now rather ancient. (Recreating is probably simpler than rebasing.)

PR #63, opened 2020-05-01
Package detail docs and minor fixes

jskeet:


jskeet:
This PR needs updating a little bit given that events have changed since I created it - but I'll merge after fixing those.

PR #75, opened 2020-05-26
Add Firebase Realtime Database event support

jskeet:
This is described at https://cloud.google.com/functions/docs/calling/realtime-database

(There are more sample JSON files than I'd normally include, but as the data isn't clearly typed, I thought it would be best to specify it like this.)

### DIFF
--- a/src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
+++ b/src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
@@ -42,6 +42,10 @@ namespace Google.Cloud.Functions.Framework.GcfEvents
             { "providers/cloud.firestore/eventTypes/document.create", "com.google.cloud.firestore.document.create.v0" },
             { "providers/cloud.firestore/eventTypes/document.update", "com.google.cloud.firestore.document.update.v0" },
             { "providers/cloud.firestore/eventTypes/document.delete", "com.google.cloud.firestore.document.delete.v0" },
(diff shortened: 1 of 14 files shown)


src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
```diff
@@ -68,9 +76,16 @@ internal static async ValueTask<CloudEvent> ConvertGcfEventToCloudEvent(HttpRequ
             {
                 throw new CloudEventConverter.ConversionException($"Unexpected event type for function: '{gcfType}'");
             }
-            if (gcfType.StartsWith("providers/cloud.firestore/eventTypes/", StringComparison.Ordinal))
+
+            // Firestore and Firebase database GCF requests include a "params" element that maps wilcards in the resource
+            // template to their value in the changed document.
+            if (gcfType.StartsWith("providers/cloud.firestore/eventTypes/", StringComparison.Ordinal) ||
+                gcfType.StartsWith("providers/google.firebase.database/", StringComparison.Ordinal))
```
amanda-tarafa, line 83: silly question: Why does the Firestore check includes the "eventTypes" but the Firebase check doesn't?

jskeet: No particular reason. Will take it out of Firestore - I can't imagine we'd get anything *other* than eventTypes after the second slash.

PR #84, opened 2020-06-11
Integrate Google.Events into Google.Functions.Framework

jskeet:
As normal, review one commit at a time.


events-src/Google.Events.SystemTextJson/Cloud/PubSub/V1/PubsubMessageData.g.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Text.Json.Serialization;
+
+// Note: we'd expect to generate this type from a JSON schema.
+
+namespace Google.Events.SystemTextJson.Cloud.PubSub.V1
+{
+    /// <summary>
+    /// The CloudEvent representation of a PubSub message.
+    /// </summary>
+    public sealed partial class PubsubMessageData
```
amanda-tarafa, line 27: Should this be called Pub**S**ubMessageData. The namespace is the file is not. There was a similar comment on the protobuf definition on yesterday's PR.

jskeet: Yup, which I would have commented on if I hadn't moved it :)
Basically, this mirrors what's in the API itself: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L188
(C# namespace includes PubSub, message is PubsubMessage.)

So I'm being consistent with the API, basically. But I agree it's annoying.

PR #106, opened 2020-07-21
Docs: Add more documentation for running examples

### LINKED ISSUE #105: Add Instructions to `examples/Google.Cloud.Functions.Examples.StorageImageAnnotator`
Add some instructions on how to run this sample:

https://github.com/GoogleCloudPlatform/functions-framework-dotnet/tree/master/examples/Google.Cloud.Functions.Examples.StorageImageAnnotator

Or perhaps some generic instructions on running samples in the `examples` folder.

jskeet:
- The README.md in examples links to the examples doc page
- The doc page gives generic instructions for running examples
- Each example now has a sample deployment command for GCF

(I'm reluctant to go any further than this, given that there will be
more complete samples on cloud.google.com for beta.)

Fixes #105.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -145,7 +145,8 @@ Sample deployment:
 ```sh
 gcloud functions deploy vb-http \
   --runtime dotnet3 \
-  --trigger-http --allow-unauthenticated \
+  --trigger-http \
+  --allow-unauthenticated \
   --entry-point Google.Cloud.Functions.Examples.VbHttpFunction.Function
 ```
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -23,6 +23,30 @@ Some examples include multiple classes in a single source file. This
 is purely to make the examples easier to follow when browsing on
 GitHub.
 
+## Running and deploying
+
+Each example can be run locally from the command line by changing to
+the example directory and running
+
+```sh
+dotnet run
+```
+
+Alternatively, you can open the solution file in the `examples`
+directory, set the "Startup Project" to the example you want to run,
+and just press F5 to run.
+
+To deploy to Google Cloud Functions or Cloud Run, follow the
+[deployment guide](deployment.md). Every example uses a function
+class called `Function` in a namespace that's the same as the
+directory. For example, the entry point for the function in the
+`Google.Cloud.Functions.Examples.SimpleHttpFuntion` directory is
+`Google.Cloud.Functions.Examples.SimpleHttpFunction.Function`.
+
+A sample command line is given for each example, but you may need to
+adapt this to specify your own Google Cloud resources such as PubSub
+topics or Storage buckets.
+
 ## SimpleHttpFunction
 
 The [SimpleHttpFunction](../examples/Google.Cloud.Functions.Examples.SimpleHttpFunction)
@@ -31,6 +55,14 @@ template using the following command line:
 
 ```sh
 dotnet new gcf-http
+```
+
+Sample deployment:
+
+```sh
+gcloud functions deploy simple-http --runtime dotnet3 \
+  --trigger-http --allow-unauthenticated \
+  --entry-point Google.Cloud.Functions.Examples.SimpleHttpFunction.Function
 ```
 
 ## SimpleEventFunction
@@ -43,6 +75,17 @@ template using the following command line:
 dotnet new gcf-event
 ```
 
+This function expects Storage events, and logs details of the event it receives.
+
+Sample deployment:
+
+```sh
+gcloud functions deploy simple-event --runtime dotnet3 \
+  --trigger-event google.storage.object.finalize \
+  --trigger-resource <gcs-bucket-name> \
+  --entry-point Google.Cloud.Functions.Examples.SimpleEventFunction.Function
+```
+
 ## SimpleUntypedEventFunction
 
 The [SimpleUntypedEventFunction](../examples/Google.Cloud.Functions.Examples.SimpleUntypedEventFunction)
@@ -51,6 +94,17 @@ template using the following command line:
 
 ```sh
 dotnet new gcf-untyped-event
+```
+
+This function can handle any event type, and logs some generic
+details of the event it receives.
+
+Sample deployment to listen to PubSub events:
+
+```sh
+gcloud functions deploy simple-untyped-event --runtime dotnet3 \
+  --trigger-topic <topic-id> \
+  --entry-point Google.Cloud.Functions.Examples.SimpleUntypedEventFunction.Function
 ```
 
 ## VbHttpFunction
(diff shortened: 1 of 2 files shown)


jskeet:
(The use of `\` for multi-line command line deployments *may* not be ideal, as it probably wouldn't work from the Windows command line - but I suspect most Functions Framework users are either running a bash-like shell, or are at least familiar enough with it to know what to do.)

Tetsuoryuu:
> (The use of `\` for multi-line command line deployments _may_ not be ideal, as it probably wouldn't work from the Windows command line - but I suspect most Functions Framework users are either running a bash-like shell, or are at least familiar enough with it to know what to do.)

I have been asked about that by one of my mentees.
Would a small note about that help? It would remove a possible pain point for someone just getting started.

jskeet:
@Tetsuoryuu: Interesting. The tricky bit is finding somewhere to put a note that isn't distracting to those who don't need it, but will be obvious to those who are trying to run the example. (I'm not convinced that everyone will have read the whole page...)

I *could* just put the whole command line on a single line in each case, of course. That would be pretty huge and (IMO) less *readable*, but make it easier to execute.

Any suggestions are most welcome!

Tetsuoryuu:
@jskeet if there were more Windows unique instructions it might make sense to make a separate doc file, but at a first glance only that one stands out, maybe it could be a one-liner at the end of the "Running and deploying" paragraph.

jskeet:
@Tetsuoryuu: I've taken a first pass at that in a second commit. (It's a bit more than a one liner, but...) If you have time to review that, it would be much appreciated!

jskeet:
@Tetsuoryuu: Wonderful - thank you so much :)

jskeet:
@grant: Ping?

docs/examples.md
```diff
@@ -23,6 +23,43 @@ Some examples include multiple classes in a single source file. This
...
+## Running and deploying
+
+Each example can be run locally from the command line by changing to
+the example directory and running
+
+```sh
+dotnet run
+```
+
+Alternatively, you can open the solution file in the `examples`
```
grant, line 35: open the solution file in...

This guide assumes I have a program to open the solution file and know what a solution file (.sln file) is.

docs/examples.md
```diff
@@ -33,6 +70,14 @@ template using the following command line:
 dotnet new gcf-http
 ```
 
+Sample deployment:
+
+```sh
+gcloud functions deploy simple-http --runtime dotnet3 \
```
grant, line 76: Nit: Separate `--runtime dotnet3` from the command with a newline like with other flags.

See Java for example:
https://cloud.google.com/functions/docs/tutorials/slack#deploying_the_function

docs/examples.md
```diff
@@ -53,6 +109,17 @@ template using the following command line:
 dotnet new gcf-untyped-event
 ```
 
+This function can handle any event type, and logs some generic
+details of the event it receives.
+
+Sample deployment to listen to PubSub events:
```
grant, line 115: Pub/Sub

docs/examples.md
```diff
@@ -63,6 +130,14 @@ template using the following command line:
 dotnet new gcf-http -lang vb
 ```
 
+Sample deployment:
+
+```sh
+gcloud functions deploy vb-http --runtime dotnet3 \
+  --trigger-http --allow-unauthenticated \
```
grant, line 137: Separate flags uniformly

docs/examples.md
```diff
@@ -43,6 +88,17 @@ template using the following command line:
...
 ```
 
+This function expects Storage events, and logs details of the event it receives.
+
+Sample deployment:
+
+```sh
+gcloud functions deploy simple-event --runtime dotnet3 \
+  --trigger-event google.storage.object.finalize \
+  --trigger-resource <gcs-bucket-name> \
```
grant, line 98: I like to use `$GCS_BUCKET_NAME` such that I can just copy/paste shell commands.

On the docs we use _`YOUR_`_ like _`YOUR_TRIGGER_BUCKET_NAME`_

https://cloud.google.com/functions/docs/tutorials/storage#object_metadata_update_triggering_the_function

Either of those versions would be slightly easier to use.

docs/examples.md
```diff
@@ -23,6 +23,43 @@ Some examples include multiple classes in a single source file. This
...
+
+To deploy to Google Cloud Functions or Cloud Run, follow the
+[deployment guide](deployment.md). Every example uses a function
+class called `Function` in a namespace that's the same as the
+directory. For example, the entry point for the function in the
+`Google.Cloud.Functions.Examples.SimpleHttpFuntion` directory is
+`Google.Cloud.Functions.Examples.SimpleHttpFunction.Function`.
+
+A sample command line is given for each example, but you may need to
+adapt this to specify your own Google Cloud resources such as PubSub
```
grant, line 47: Pub/Sub

docs/examples.md
jskeet: Yes, it does - because I'd expect every .NET developer to have that information already. I'm very reluctant to reduce the information density by adding extra text that is irrelevant to 99.9% of the expected audience.

docs/examples.md
jskeet: Done (here and elsewhere). I'll do a quick grep to see if I've got this elsewhere other than identifiers - but I'll make changes to other files in a new PR.

docs/examples.md
jskeet: Done, although it's not clear to me why the name of the function (which is another flag, effectively- just an unnamed one) should stay on the first line.
Personally I preferred it as it was, but I'm not going to argue over it.

docs/examples.md
jskeet: We can't use italics in Markdown code blocks. I've used GCS_BUCKET_NAME, but I've also added an explanatory comment at the start to indicate that you don't *need* those environment variables. I dislike guides that tell you to set environment variables but don't explain that it's just for the sake of the command lines shown - it leaves readers potentially confused about whether the tools themselves will use the environment variables (as they do for GOOGLE_APPLICATION_CREDENTIALS for example).

docs/examples.md
jskeet: Done

docs/examples.md
```diff
@@ -85,6 +170,17 @@ template using the following command line:
 dotnet new gcf-untyped-event -lang vb
 ```
 
+This function can handle any event type, and logs some generic
+details of the event it receives.
+
+Sample deployment to listen to PubSub events:
```
jskeet, line 176: Fixed here too.

docs/examples.md
```diff
@@ -115,6 +230,17 @@ template using the following command line:
 dotnet new gcf-untyped-event -lang f#
 ```
 
+This function can handle any event type, and logs some generic
+details of the event it receives.
+
+Sample deployment to listen to PubSub events:
```
jskeet, line 236: And here

grant:
Needs changes pushed. I can't see changes

jskeet:
Doh - I'd run "git push" but hadn't spotted it had failed. It's there now.

docs/examples.md
jskeet: Missed this one. On it.

jskeet:
Merging this now; we can discuss the one remaining issue about whether or not more detail is needed about solution files at another time if necessary.

PR #110, opened 2020-07-27
Fix heartbeat warning

jskeet:
(First commit is just refactoring; second commit is just about heartbeats.)

### DIFF
--- a/src/Google.Cloud.Functions.Invoker/Logging/LoggerBase.cs
+++ b/src/Google.Cloud.Functions.Invoker/Logging/LoggerBase.cs
@@ -22,10 +22,17 @@ namespace Google.Cloud.Functions.Invoker.Logging
     /// </summary>
     internal abstract class LoggerBase : ILogger
     {
+        private const string KestrelCategory = "Microsoft.AspNetCore.Server.Kestrel";
+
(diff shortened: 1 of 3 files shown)


src/Google.Cloud.Functions.Invoker/Logging/LoggerBase.cs
```diff
@@ -39,6 +46,12 @@ internal abstract class LoggerBase : ILogger
         /// </summary>
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
         {
+            // Functions expect to go for a long time without any CPU. It's reasonable to suppress this warning.
+            if (_isKestrelCategory && eventId.Id == HeartbeatSlowEventId)
```
amanda-tarafa, line 50: Would it be possible to make this configurable? Even if we suppress it by default...
My heart is not set on it though.

jskeet: Getting configuration to the logger would be fairly tortuous, I think. Let's leave it as unconfigurable for now, and wait for the feature request :)

PR #112, opened 2020-07-29
Add a FunctionTestBase to derive from in integration tests

jskeet:
Writing the samples has proved this to be extremely convenient.

Note: I'm very much up for a debate about the methods to include around making requests. These aren't quite like the ones in the samples - in particular, I never return HttpResponseMessage, because returning that after disposing of the client could do nasty things in some cases. (That's why I have the action for validation.)

### DIFF
--- a/src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
+++ b/src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
@@ -1,0 +1,125 @@
+﻿// Copyright 2020, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using CloudNative.CloudEvents;
+using Microsoft.AspNetCore.Authorization.Infrastructure;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Net.Mime;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Invoker.Testing
+{
+    /// <summary>
+    /// Base class for test for functions. This class is designed
+    /// for use in test frameworks that dispose of test classes
+    /// automatically. (This includes NUnit and xUnit.) This class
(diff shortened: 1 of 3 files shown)


jskeet:
Thanks Amanda - I'll add docs in another commit before merging, and we might want to have a quick chat about the method signatures etc.

src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
```diff
@@ -0,0 +1,124 @@
...
+            using var response = await client.SendAsync(request);
+            validator(response);
+        }
+
+        /// <summary>
+        /// Executes the function with the specified CloudEvent, which is converted
+        /// to an HTTP POST request using structured encoding. This method asserts that the
+        /// request completed successfully.
+        /// </summary>
+        public async Task ExecuteCloudEventRequestAsync(CloudEvent cloudEvent)
```
amanda-tarafa, line 93: Consider adding an overload taking a validator. We might not need one now, but maybe in the future, and we don't need to release this package just because of that, and also for completeness.

src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
```diff
@@ -0,0 +1,124 @@
...
+            var response = await client.SendAsync(request);
+            response.EnsureSuccessStatusCode();
+        }
+
+        /// <summary>
+        /// Returns a list of the log entries for the category associated with the function.
+        /// </summary>
+        /// <returns>A list of log entries. The returned value may refer to an empty list,
+        /// but is never null.</returns>
+        public List<TestLogEntry> GetFunctionLogEntries() => Server.GetFunctionLogEntries();
```
amanda-tarafa, line 117: Given that you are adding this one helper here, consider adding another for Server.ClearLogs().

src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
```diff
@@ -0,0 +1,124 @@
...
+        protected FunctionTestBase() : this(new FunctionTestServer<TFunction>())
+        {
+        }
+
+        /// <summary>
+        /// Executes a simple HTTP GET request against an arbitrary URI, and returns
+        /// the text of the response after validating that it indicates success. For
+        /// more complex scenarios, use <see cref="ExecuteHttpRequestAsync(HttpRequestMessage, Action{HttpResponseMessage})"/>.
+        /// </summary>
+        public async Task<string> ExecuteHttpGetRequestAsync()
```
amanda-tarafa, line 70: Consider an optional uri parameter. There might be functions that validate the calling uri I suppose.

src/Google.Cloud.Functions.Invoker.Testing/FunctionTestBase.cs
amanda-tarafa, line 93: Consider an optional uri parameter.

jskeet:
Merging now, but I need to add more docs later.

PR #114, opened 2020-07-31
docs: Add more testing docs for logging and FunctionTestBase

jskeet:


docs/testing.md
```diff
@@ -92,3 +133,21 @@ the demands of most users. Aspects that may change later include:
...
+## Testing logs in unit tests
+
+If you want to write a unit test for a function whose constructor
+accepts an `ILogger<T>`, you can use the built-in ASP.NET Core
+`NullLogger<T>` if you don't need to test the log entries. However,
+this doesn't help if you want to make sure that the expected log
+entries are emitted. The logger implementation used by
+`FunctionTestServer` is also available for standalone testing. Just
+create an instance of
+[FunctionTestServer&lt;TCategoryName>](../src/Google.Cloud.Functions.Invoker.Testing/MemoryLogger.cs)
```
amanda-tarafa, line 146: nit: the less than is encoded but the greater than is not, I suppose the less than can be unencoded because it's been used throught the docs and it shows well.

jskeet: I'd expect it to be fine within a code block, but not directly - otherwise Markdown will think it's a `<TCategoryName>`  HTML tag. Just tried it in this comment, and we really do need it to be encoded.

amanda-tarafa: Ahhh, I see, and you only encode the less than because then Markdown won't try to interpret the >. Got it.

Issue #115, opened 2020-08-03
Use GitHub Actions Rather than Travis


grant:
For various reasons, we should use GitHub actions for CI rather than Travis:

## Expected

- Use GitHub Actions
- No Travis config: https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/master/.travis.yml

---

Here's an example GitHub Action workflow we can use:
https://github.com/actions/setup-dotnet

```yaml
steps:
- uses: actions/checkout@main
- uses: actions/setup-dotnet@v1
  with:
    dotnet-version: '3.1.x' # SDK Version to use; x will use the latest version of the 3.1 channel
- run: ./build.sh
```

jskeet:
(Discussing internally.)

jskeet:
Fixed in #116.

PR #116, opened 2020-08-04
Move to GitHub Actions for build and test

jskeet:


jskeet:
Hmm. Need to work out why this hasn't triggered. I wonder whether it needs merging before running for the first time. Will experiment in a personal repo.

grant:
Ah, forgot about this PR.
Here's the issue https://github.com/GoogleCloudPlatform/functions-framework-dotnet/issues/115

.github/workflows/build.yml
```diff
@@ -0,0 +1,17 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.x
+    - run: ./build.sh
```
grant, line 17: Personally, it would be nice to just put the full build action in the workflow rather than a shell script. We'll get consistent results and it shows up nicely in the UI rather than mixing the two.

Although the downside is that you can't run a workflow locally.

What do other Yoshi repos do? I'd probably say fully using the action and removing the `build.sh` file would be good and we can just create branches/PRs if we'd like to run the build.

Either way, looks good.

jskeet:
Ha - I need to change the GitHub settings to not require the Travis status before I can merge this... I'll then re-enable it with the GitHub actions status instead.

PR #117, opened 2020-08-05
Rearchitecture of Functions Framework hosting

jskeet:
Tests cover the same scenarios as before (and a few more).

This commit does not include documentation, or samples of additional
functionality (e.g. FunctionTestServerBuilder). However, it is
complete in itself.

I expect the review to primarily cover naming, and what the extension methods should target.

(Once that has been decided with @amanda-tarafa, I'll invite additional reviewers.)


src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,159 @@
...
+                HostingInternals.AddServicesForFunctionTarget(serviceCollection, HostingInternals.GetFunctionTarget(context, assembly)));
+
+        /// <summary>
+        /// Configures the services within the given web host builder for the Functions Framework to use
+        /// the function target type specified by the <typeparamref name="TFunction"/> type parameter.
+        /// </summary>
+        /// <typeparam name="TFunction">The function target type.</typeparam>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <returns>The original builder, for method chaining.</returns>
+        public static IWebHostBuilder AddFunctionTarget<TFunction>(this IWebHostBuilder builder) =>
```
jskeet, line 109: I've been thinking about this a bit more, and it probably does make more sense for this to be on IServiceCollection. Let me know if you agree, and I'll amend the PR.

jskeet: (This is now done.)

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// </summary>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <param name="startup">The startup class to use.</param>
+        /// <returns>The original builder, for method chaining.</returns>
+        public static IWebHostBuilder AddFunctionsStartup(this IWebHostBuilder builder, FunctionsStartup startup) =>
+            builder.ConfigureServices((context, services) => HostingInternals.AddStartup(context, services, startup));
+
+        /// <summary>
+        /// Configures the given web host builder to use the Functions Framework.
+        /// This method executes any application configurers before adding handlers for fixed paths to
```
amanda-tarafa, line 166:
```suggestion
        /// This method executes any application Startups before adding handlers to
```

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// return "not found" responses for fixed paths (e.g. "favicon.ico") and setting the terminal
+        /// handler to execute the target function.
+        /// </summary>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <returns>The original builder, for method chaining.</returns>
+        public static IWebHostBuilder UseFunctionsFramework(this IWebHostBuilder builder) =>
```
amanda-tarafa, line 172: I think I would add some remarks saying all the other things that **need** to be configured before calling this one, I think at least the configuration source and the target function. People might get the wrong idea that this does everything with some defaults.

src/Google.Cloud.Functions.Invoker/HostingInternals.cs
```diff
@@ -0,0 +1,203 @@
...
+            assembly
+                .GetCustomAttributes<FunctionsStartupAttribute>()
+                .OrderBy(attr => attr.Order)
+                .ThenBy(attr => attr.StartupType.FullName)
+                .Select(attr => Activator.CreateInstance(attr.StartupType))
+                .Cast<FunctionsStartup>();
+
+        internal static void AddStartup(WebHostBuilderContext context, IServiceCollection services, FunctionsStartup startup)
+        {
+            startup.ConfigureServices(context, services);
```
amanda-tarafa, line 89: I think this should be done in ConfigureApplication. The user might be trying to override some services and when doing it here, it's us who might end overriding theirs. We "promised" to execute the Startups at the end, on the UseFunctionsFramework extension, but we are setting the services right here instead. Else add a note on the public extensions that add startups that the services will be added inmediately but the midlewares will be added later.

src/Google.Cloud.Functions.Invoker/HostingInternals.cs
```diff
@@ -0,0 +1,203 @@
...
+            // Remember the startup for application configuration time as well.
+            services.AddSingleton(startup);
+        }
+
+        private static void ReturnNotFound(IApplicationBuilder app) =>
```
amanda-tarafa, line 94: Consider adding this one as a local method on ConfigureApplication.

src/Google.Cloud.Functions.Invoker/HostingInternals.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                context.Response.StatusCode = (int) HttpStatusCode.NotFound;
+                return Task.CompletedTask;
+            });
+
+        /// <summary>
+        /// Executes a request by asking the dependency injection context for an IHttpFunction, then
+        /// executing it.
+        /// </summary>
+        internal static async Task Execute(HttpContext context)
```
amanda-tarafa, line 105: And this one as well, maybe belongs on ConfigureApplication

src/Google.Cloud.Functions.Invoker/FunctionsStartup.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Google.Cloud.Functions.Invoker
+{
+    /// <summary>
+    /// Base class for startup classes providing additional configuration for functions
+    /// in the Functions Framework, typically configuring services and/or middleware.
+    /// These types are discovered using <see cref="FunctionsStartupAttribute"/>,
+    /// which also allows ordering to be specified.
```
amanda-tarafa, line 25: This is partially true only. We can ensure that services in the startups will be included in the service collection in that order but not middlewares (where the order is also important), because although we add the Startup classes to the services in the order indicated by the attribute, we have no guarante that when we get them on ConfigureApplication we'll get them on that order. I think you'll need a wrapper for the Startups that allows you to add an order both based on the attribute and on the actual order the single ones were included, so you can mimick that later on ConfigureApplication.
Or lets chat because it'll be easier to explain.

src/Google.Cloud.Functions.Invoker/FunctionsStartupAttribute.cs
```diff
@@ -14,25 +14,33 @@
...
 using System;
 
-namespace Google.Cloud.Functions.Invoker.DependencyInjection
+namespace Google.Cloud.Functions.Invoker
 {
     /// <summary>
-    /// Attribute used to declare <see cref="FunctionsStartup"/> types, primarily used to
-    /// configure dependency injection within functions. When multiple types are specified,
-    /// the order in which they are used is undefined.
+    /// Assembly attribute to specify a <see cref="FunctionsStartup"/> to use
```
amanda-tarafa, line 20:
```suggestion
    /// Assembly attribute to specify one or many <see cref="FunctionsStartup"/> to use
```

src/Google.Cloud.Functions.Invoker.Tests/FunctionsStartupTest.cs
```diff
@@ -0,0 +1,185 @@
...
+        public class TestStartup2 : FunctionsStartup { }
+
+        // This is the only test that uses the assembly attribute detection.
+        // Everything else specifies the startups explicitly.
+        [Fact]
+        public void GetStartups()
+        {
+            var startups = HostingInternals.GetStartups(typeof(FunctionsStartupTest).Assembly);
+            var actualTypes = startups.Select(startup => startup.GetType()).ToArray();
+            // TestStartup2 comes before TestStartup1 due to the Order properties.
```
amanda-tarafa, line 49: This is what we can't ensure after we've added them to the Services Collection, getting them for executing them App Configuration might give them out in a different order.

jskeet: https://www.stevejgordon.co.uk/asp-net-core-dependency-injection-registering-multiple-implementations-interface includes:

> It’s worth making it clear that the implementations will be added in the order they are registered. They will be returned in that same order when injected into calling code. Depending on your requirements, this may be useful and important.

That's not official documentation, but it's at least encouraging.

src/Google.Cloud.Functions.Invoker/FunctionsEnvironmentVariablesConfigurationSource.cs
```diff
@@ -0,0 +1,75 @@
...
+                MaybeAdd(EntryPoint.FunctionTargetEnvironmentVariable, nameof(FunctionsFrameworkOptions.FunctionTarget), target => target);
+                MaybeAdd(EntryPoint.PortEnvironmentVariable, "Port", port => port);
+                MaybeAdd("DOTNET_RUNNING_IN_CONTAINER", "Address",
+                    value => string.Equals(value, "true", StringComparison.OrdinalIgnoreCase) ? "Any" : "Loopback");
+                MaybeAdd("K_SERVICE", "JsonLogging", service => string.IsNullOrWhiteSpace(service) ? "False" : "True");
+
+                // FUNCTION_TARGET => Target
+                // PORT => Port
+                // DOTNET_RUNNING_IN_CONTAINER (true?) => Address = "Any" / "Loopback"
+                // K_SERVICE (any) => JsonLogging = True, False
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <returns>The original builder, for method chaining.</returns>
+        public static IWebHostBuilder ConfigureKestrelForFunctionsFramework(this IWebHostBuilder builder) =>
+            builder.ConfigureKestrel((context, serverOptions) =>
+            {
+                var options = FunctionsFrameworkOptions.FromConfiguration(context.Configuration);
+                serverOptions.Listen(options.GetIPAddress(), options.Port);
+            });
+
+        /// <summary>
+        /// Configures the services within the given web host builder for the Functions Framework to use
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// If this method completes successfully, a binding for <see cref="IHttpFunction"/> will definitely
+        /// have been added to the service collection. Other bindings may also be present, in order to adapt
+        /// the function to <see cref="IHttpFunction"/>.
+        /// </remarks>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <param name="assembly">The assembly expected to contain the Functions Framework target function.</param>
+        /// <returns>The original builder, for method chaining.</returns>
+        public static IWebHostBuilder AddFunctionTarget(this IWebHostBuilder builder, Assembly assembly) =>
+            builder.ConfigureServices((context, serviceCollection) =>
+                HostingInternals.AddServicesForFunctionTarget(serviceCollection, HostingInternals.GetFunctionTarget(context, assembly)));
```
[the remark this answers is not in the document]
jskeet: Ah yes, that's nicer.

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// the function to <see cref="IHttpFunction"/>.
+        /// </remarks>
+        /// <typeparam name="TFunction">The function target type.</typeparam>
+        /// <param name="services">The service collection to configure.</param>
+        /// <returns>The original service collection, for method chaining.</returns>
+        public static IServiceCollection AddFunctionTarget<TFunction>(this IServiceCollection services) =>
+            services.AddFunctionTarget(typeof(TFunction));
+
+        /// <summary>
+        /// Configures the given service collection to use the specified target function type, which must
```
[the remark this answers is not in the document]
jskeet: Done

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
jskeet: Doh, thanks.

jskeet: Still unsure whether these should actually be on IServiceCollection - if we accept that they *are* services, that both configure services *and take part in app configuration* then maybe it's okay to put the extension there?

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
jskeet: Good plan. Alternatively, we could make *this* method do "the lot" and put a separate extension method on IApplicationBuilder.

I believe that's what you suggested in #118, and I dismissed it at the time - but I've thought of a good use case now, which is "I want to add a new configuration provider, but otherwise use normal startup".

It's possible that having EntryPoint methods accepting configuration delegates for that would cover that. Let's talk about it more before merging - I think I'm going to write docs with sample use cases before merging too, to make sure everything is feasible.

src/Google.Cloud.Functions.Invoker/FunctionsStartup.cs
jskeet: Covered by earlier comment, I believe - I think the order *is* preserved. We could potentially expand the comment to talk about ordering with respect to explicitly-added startups, but it might add more noise than enlightenment.

Definitely happy to chat.

src/Google.Cloud.Functions.Invoker/FunctionsStartupAttribute.cs
jskeet: No, this attribute (as in an instance of it) only specifies a single startup type - but it can be applied multiple times. Will add that as another sentence though.

src/Google.Cloud.Functions.Invoker/FunctionsExtensions.cs
jskeet: The more I read the comments on the methods, the more it feels like they should be on IServiceCollection. Doing that now, and you can argue against it later if you want :)

src/Google.Cloud.Functions.Invoker/Extensions/FunctionsFrameworkLoggingExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Functions.Invoker;
+using Google.Cloud.Functions.Invoker.Logging;
+using Microsoft.AspNetCore.Hosting;
+
+namespace Microsoft.Extensions.Logging
```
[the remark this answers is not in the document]
jskeet: Yup, I've actually got that locally, but didn't want to disrupt the review by pushing while you were reviewing.

src/Google.Cloud.Functions.Invoker/Extensions/FunctionsFrameworkLoggingExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Extensions for adding configuration for the Functions Framework.
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Invoker/Extensions/FunctionsFrameworkWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                var options = FunctionsFrameworkOptions.FromConfiguration(context.Configuration);
+                serverOptions.Listen(options.GetIPAddress(), options.Port);
+            });
+
+        /// <summary>
+        /// Configures the given web host builder to use the Functions Framework.
+        /// This method executes the <see cref="FunctionsStartup.Configure(WebHostBuilderContext, AspNetCore.Builder.IApplicationBuilder)"/>
+        /// method on any registered functions startup classes before adding handlers for fixed paths to
+        /// return "not found" responses for fixed paths (e.g. "favicon.ico") and setting the terminal
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Invoker.Tests/FunctionTargetTest.cs
```diff
@@ -0,0 +1,236 @@
...
+            Assert.Throws<InvalidOperationException>(() => HostingInternals.AddServicesForFunctionTarget(services, typeof(NonFunction)));
+        }
+
+        [Fact]
+        public void NonInstantiableFunctionType()
+        {
+            var services = new ServiceCollection();
+            services.AddLogging();
+            HostingInternals.AddServicesForFunctionTarget(services, typeof(NonInstantiableFunction));
+            // We only find out when we try to get an IHttpFunction that we can't actually instantiate the function.
```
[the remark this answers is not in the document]
jskeet: That might do odd things, as we'd be requesting a normally-request-oriented dependency not within the scope of a request.
Normally that should be fine, but I wouldn't be *entirely* surprised to see it break something.

src/Google.Cloud.Functions.Invoker.Tests/FunctionTargetTest.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        [Fact]
+        public void FindDefaultFunctionType_SingleFunctionType()
+        {
+            var expected = typeof(DefaultFunction);
+            var actual = HostingInternals.FindDefaultFunctionType(
+                typeof(FunctionTargetTest),
+                typeof(FunctionsEnvironmentVariablesConfigurationSourceTest),
+                typeof(AbstractHttpFunction), // Abstract, doesn't count
+                typeof(INamedHttpFunction),    // Interface, doesn't count
```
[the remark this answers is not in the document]
jskeet: Fixed, and below.

src/Google.Cloud.Functions.Invoker/HostingInternals.cs
jskeet: Ooh, missed this comment. Will do.

src/Google.Cloud.Functions.Invoker/Extensions/FunctionsFrameworkWebHostBuilderExtensions.cs
```diff
@@ -41,35 +41,32 @@ public static class FunctionsFrameworkWebHostBuilderExtensions
...
-        /// <item><description>Kestrel is configured with the address and port configured from the environment and command line.</description></item>
-        /// <item><description>The function target is located within <paramref name="functionAssembly"/> based on configuration, or the function assembly
-        /// is scanned for a single function if the configuration does not specify a function target.
-        /// </description></item>
-        /// <item><description>Function startup classes (typically used to configure the services for dependency injection or to add middleware) are
-        /// added based on any <see cref="FunctionsStartupAttribute"/> values applied to the function assembly.
-        /// <item><description>The application is configured to use the Functions Framework request handler.</description></item>
-        /// </description></item>
-        /// </list>
+        /// Uses the functions startup classes specified by assembly attributes when configuring the host.
```
amanda-tarafa, line 44: nit: Maybe add a <see href.... /> for the attribute type.

src/Google.Cloud.Functions.Invoker/EntryPoint.cs
```diff
@@ -60,7 +60,14 @@ public static async Task<int> StartAsync(Assembly functionAssembly, string[] arg
             // doesn't matter much. Potentially catch exceptions during configuration, but let any
             // during web server execution propagate.
             var host = Host.CreateDefaultBuilder()
-                .ConfigureWebHostDefaults(webHostBuilder => webHostBuilder.UseFunctionsFramework(functionAssembly, args))
+                .ConfigureWebHostDefaults(webHostBuilder => webHostBuilder
+                    .ConfigureAppConfiguration(builder => builder.AddFunctionsEnvironment().AddFunctionsCommandLine(args))
+                    .ConfigureLogging((context, logging) => logging.ClearProviders().AddFunctionsConsoleLogging(context))
+                    .ConfigureKestrelForFunctionsFramework()
+                    .ConfigureServices((context, services) => services
+                    .AddFunctionTarget(context, functionAssembly))
```
amanda-tarafa, line 68: Consider creating our own DefaultStartup and adding all of these in there (except maybe the Kestrel bit) and then calling here .UseFunctionStartup(new DefaultStartup()) just before .UserFunctionsStartups(functionAssembly).

src/Google.Cloud.Functions.Invoker/Extensions/FunctionsFrameworkWebHostBuilderExtensions.cs
jskeet: Done.

src/Google.Cloud.Functions.Invoker/EntryPoint.cs
jskeet: Maybe. That has a certain elegance to it, but it possibly obscures what we're doing a little. Let's leave that as an idea for the future.

jskeet:
Docs are now ready for review, @amanda-tarafa - just the fourth commit, basically.

Once you're happy, we can invite others to look :)

docs/customization.md
```diff
@@ -0,0 +1,167 @@
...
+In more complex scenarios, users may wish to provide additional
+dependencies themselves through Functions Startup classes, by
+overriding the `ConfigureServices` method:
+
+```csharp
+public virtual void ConfigureServices(WebHostBuilderContext context, IServiceCollection services)
+```
+
+The dependencies are then applied by ASP.NET Core dependency
+injection in the normal way. The
```
amanda-tarafa, line 105:
```suggestion
injection in the normal way. The ???
```

docs/customization.md
```diff
@@ -0,0 +1,167 @@
...
+public virtual void Configure(WebHostBuilderContext context, IApplicationBuilder app)
+```
+
+The code in the Functions Startup is run during application
+configuration, before the Functions Framework configures its
+own request handling. This allows you to add additional middleware.
+Note that while nothing prevents you from installing arbitrary
+middleware that terminates the pipeline early, the expectation is
+that successful requests will always end up reaching the function.
+(For example, installing middleware that terminates the request with
```
amanda-tarafa, line 128: This is also the expected behaviour for "normal" ASP.NET Core Application Middleware, so it might make it cleare to add something like "As with any other ASP.NET Core app, installing middleware...." Mostly to make it clearer that this is not realy a a Functions Framework thing.

docs/customization.md
```diff
@@ -0,0 +1,167 @@
...
+an error for certain conditions is fine; installing middleware to
+respond with files from the file system is less conventional.)
+
+See [Middleware.Function](../examples/Google.Cloud.Functions.Examples.Middleware/Function.cs)
+for an example of a small piece of middleware providing extra per-request logging.
+
+## Customizing application configuration sources using `ConfigureAppConfiguration`
```
amanda-tarafa, line 135: I suppose you explicitly orderd these sections by importance or provability of use. I think that's fine, just calling out here that maybe a note on execution order is relevant.
For instance, it's relevant that configuration added here is available for everything else, etc.

docs/customization.md
jskeet: Hmm. That's odd. Not sure what I was trying to say there - have removed that sentence and brought the example into the same paragraph.

docs/customization.md
jskeet: No, it really *is* a FF thing. Various bits of middleware *will* terminate the pipeline successfully early, e.g. by serving a static file. A function really shouldn't be trying to act as a regular web server - it should be about executing the function.

docs/customization.md
jskeet: Good plan, done.

jskeet:
+Julie and Huy for optional review and a general heads-up.

jskeet:
Merging this now to unblock other work, but feedback is still welcome.

PR #118, opened 2020-08-05
Alternate extension methods

amanda-tarafa:
I believe this matches better with ASP.NET Core conventions.
Changes in HostingExtensions and EntryPoint only, with respect to #117 (didn't want to commit there so as not to mess your branch).
And I realize I created a branch on this repo, I'll delete after.


src/Google.Cloud.Functions.Invoker/EnvironmentVariablesConfigurationSource.cs
```diff
@@ -22,17 +22,17 @@ namespace Google.Cloud.Functions.Invoker
     /// An IConfigurationSource that deals with *specific* Functions Framework environment variables
     /// (unlike the more general EnvironmentVariablesConfigurationSource).
     /// </summary>
-    internal sealed class FunctionsEnvironmentVariablesConfigurationSource : IConfigurationSource
+    internal sealed class FunctionsEnvironmentVariablesConfigurationProvider : IConfigurationSource
```
jskeet, line 25: I'd rather keep calling this a source given that it implements IConfigurationSource instead of IConfigurationProvider.

src/Google.Cloud.Functions.Invoker/HostingExtensions.cs
```diff
@@ -18,30 +20,24 @@ namespace Google.Cloud.Functions.Invoker
...
-        /// Adds Functions Framework configuration based on environment variables (e.g. PORT and FUNCTION_TARGET).
+        /// Adds a Functions Framework configuration provider based on environment variables (e.g. PORT and FUNCTION_TARGET).
         /// </summary>
-        /// <param name="builder">The web host builder to configure.</param>
+        /// <param name="configBuilder">The configuration builder to add the provider to.</param>
         /// <returns>The original builder, for method chaining.</returns>
-        public static IWebHostBuilder AddFunctionsFrameworkEnvironmentConfiguration(this IWebHostBuilder builder) =>
-            builder.ConfigureAppConfiguration(configBuilder => configBuilder.Add(new FunctionsEnvironmentVariablesConfigurationSource()));
+        public static IConfigurationBuilder AddFunctionsFrameworkEnvironment(this IConfigurationBuilder configBuilder) =>
+            configBuilder.Add(new FunctionsEnvironmentVariablesConfigurationProvider());
```
jskeet, line 28: Commenting as I go... these two make sense to me.

src/Google.Cloud.Functions.Invoker/HostingExtensions.cs
```diff
@@ -50,7 +46,7 @@ public static class HostingExtensions
         /// </summary>
         /// <param name="builder">The logging builder to configure.</param>
         /// <returns>The original builder, for method chaining.</returns>
-        public static ILoggingBuilder ConfigureFunctionsFrameworkLogging(this ILoggingBuilder builder, WebHostBuilderContext context) =>
+        public static ILoggingBuilder AddFunctionsFrameworkLogging(this ILoggingBuilder builder, WebHostBuilderContext context) =>
```
jskeet, line 49: This was deliberately Configure rather than Add because it does more than add - it clears existing providers.

src/Google.Cloud.Functions.Invoker/HostingExtensions.cs
```diff
@@ -82,87 +78,128 @@ public static ILoggingBuilder AddFunctionsFrameworkConsoleLogging(this ILoggingB
...
         /// </summary>
-        /// <param name="builder">The web host builder to configure.</param>
+        /// <param name="services">The service collection to add the service to.</param>
         /// <param name="assembly">The assembly expected to contain the Functions Framework target function.</param>
-        /// <returns>The original builder, for method chaining.</returns>
-        public static IWebHostBuilder ConfigureFunctionsFrameworkTarget(this IWebHostBuilder builder, Assembly assembly) =>
-            builder.ConfigureServices((context, serviceCollection) =>
-                HostingInternals.AddServicesForTarget(serviceCollection, HostingInternals.GetFunctionTarget(context, assembly)));
+        /// <returns>The original service collection, for method chaining.</returns>
+        public static IServiceCollection AddFunctionsFrameworkTarget(this IServiceCollection services, WebHostBuilderContext context, Assembly assembly)
```
jskeet, line 89: Not sure about this one - Add sounds like you might be able to use it multiple times, but that's not the case. Will defer for now.
In terms of using IServiceCollection, I'm not sure - it makes it a bit harder to do this elsewhere.

(Same will go for all the methods.)

src/Google.Cloud.Functions.Invoker/HostingExtensions.cs
```diff
@@ -82,87 +78,128 @@ public static ILoggingBuilder AddFunctionsFrameworkConsoleLogging(this ILoggingB
...
+        }
+
+        /// <summary>
+        /// Helper to configure Functions Framework with default settings.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="functionAssembly"></param>
+        /// <param name="args"></param>
+        /// <returns></returns>
+        public static IWebHostBuilder UseFunctionsFramework(this IWebHostBuilder builder, Assembly functionAssembly, string[] args)
```
jskeet, line 185: I originally considered having a method like this, but then thought that the only time you want to configure things manually is when you want some of this not to occur - in which case you wouldn't want to call the method.

The use cases for users calling any of these methods themselves are primarily if they're customizing the pipeline from scratch.

The other use cases is in FunctionTestServer Builder:)

jskeet:
Closing this now as we've discussed everything here.

PR #121, opened 2020-08-20
ci: init github actions ci

grant:
Inits GitHub Actions (if we want to go that way). Fixes https://github.com/GoogleCloudPlatform/functions-framework-dotnet/issues/115

(Pardon the multiple commits)

---

Test run:
https://github.com/grant/functions-framework-dotnet/runs/1006107018?check_suite_focus=true

Screen Shot 2020-08-20 at 10 29 47

Not sure what your thoughts on, especially with using GitHub Actions vs shell scripts.


jskeet (commented):
Why not just run build.sh? We need that anyway, and I'd rather not have duplication.

I've already got that ready to go in https://github.com/GoogleCloudPlatform/functions-framework-dotnet/pull/116 - I was just waiting for confirmation that GitHub Actions was the direction of travel before merging.

grant:
Went with #116.

PR #123, opened 2020-08-25
Update events packages

jskeet:
(Review commit-wise.)

### DIFF
--- a/src/Google.Cloud.Functions.Framework.Tests/GcfEvents/EventDeserializationTest.cs
+++ b/src/Google.Cloud.Functions.Framework.Tests/GcfEvents/EventDeserializationTest.cs
@@ -14,16 +14,17 @@
 
 using Google.Cloud.Functions.Framework.GcfEvents;
 using Google.Events;
-using Google.Events.SystemTextJson.Cloud.PubSub.V1;
-using Google.Events.SystemTextJson.Cloud.Storage.V1;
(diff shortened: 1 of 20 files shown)


src/Google.Cloud.Functions.Framework.Tests/GcfEvents/EventDeserializationTest.cs
```diff
@@ -144,45 +140,38 @@ public async Task LegacyEvents()
...
 
         [Fact]
         public async Task FirebaseDatabaseEvent_NullData()
         {
-            var firebaseEvent = await ConvertAndDeserialize<DocumentEventData>("firebase-db1.json");
-            Assert.Null(firebaseEvent.Data);
-        }
-
-        [Fact]
-        public async Task FirebaseDatabaseEvent_Wildcards()
```
amanda-tarafa, line 157: I suppose wildcards are not special anymore, right?

amanda-tarafa (approved):
LGTM

src/Google.Cloud.Functions.Framework.Tests/GcfEvents/EventDeserializationTest.cs
jskeet: They're not part of the event payload. (They were in an odd place in the GCF event, and we've decided not to include them in the CloudEvent representation.)

PR #136, opened 2020-09-21
ci: use ci matrix

grant:
- Uses very specific workflow name "[LANG] Unit CI" with appropriate filename
- Uses a build matrix (like other FFs) to support multiple versions/platforms
- Doesn't really touch anything else.

---

Note: This is the same workflow formatting across FFs.

Context:
See https://github.com/GoogleCloudPlatform/functions-framework/pull/37


jskeet (approved):
I'd personally prefer to use a *specific* version of each OS instead of "latest", but I'm *okay* with this as-is.

jskeet:
Will change settings temporarily so I can merge - otherwise the fact that `build` has been deleted will prevent it from being merged...

jskeet:
(Changed the settings permanently to look for "test" instead of "build".)

PR #143, opened 2020-09-25
Include Hosting package in Middleware example

jskeet:
The build.sh script didn't detect the problem because we always
build with LocalFunctionsFramework set to true - which makes sense
in terms of allowing us to write examples for unreleased framework
features, but hides this sort of thing. I think it's a pretty rare
problem - I don't think we need to change the build.


amanda-tarafa (approved):
LGTM.
Agree this won't happen often.

PR #146, opened 2020-09-30
docs: Add an example of deploying a multi-project function

jskeet:
Here MultiProjectFunction depends (locally) on
MultiProjectDependency, so both projects must be uploaded together,
and the correct project specified using a build-time environment
variable.

PR #147, opened 2020-09-30
Conditionally generate entry point

jskeet:
This PR builds on #146 - there will only be a single commit once that's merged.


src/Google.Cloud.Functions.Hosting/targets/Google.Cloud.Functions.Hosting.targets
```diff
@@ -29,7 +29,7 @@
           Outputs="$(_GeneratedEntryPointFile)">
     <PropertyGroup Condition="'$(Language)' == 'C#'">
       <_GeneratedProgramFileContent>
-<![CDATA[
+        <![CDATA[
```
jskeet, line 32: Note that this just formats this line like the ones later on for VB and F#

jskeet:
Rebased, so it's not just the one commit.

amanda-tarafa (approved):
LGTM

PR #148, opened 2020-09-30
ci: add dotnet format tool for linting

grant:
- Adds dotnet format tool CI workflow
- Runs format tool

Hoping to add another green box on https://github.com/GoogleCloudPlatform/functions-framework

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+root = true
+
+[*.cs]
+csharp_space_after_cast = true
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,18 @@
+name: .NET Lint CI
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+jobs:
+  lint:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v2
+      with:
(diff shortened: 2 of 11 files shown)


jskeet:
I'd *at least* want to change the settings to have a space around casts. It's not *accidental* that they're there - I find them more readable that way.

I'm fine losing the whitespace lines - but this does feel like work that isn't actually adding any customer value.

grant:
I'm indifferent about the specific formatting rules and was just using the defaults.

Is there a config we want to use with `dotnet-format` other than the default?

jskeet:
> Is there a config we want to use with dotnet-format other than the default?

I don't know, because I've never looked into it - it's simply never been something we've pursued.

grant:
I updated the editor config to have a space after cast. Other configs would be default.

Here's an example config:
https://github.com/dotnet/format/blob/master/tests/projects/for_code_formatter/formatted_project/.editorconfig

jskeet (approved):
Given that you've done all the work for this, I'm happy enough for this to go in, which isn't to say it's something I'd want to do elsewhere.

grant:
Oh I spent <20 minutes on this. Was just talking with Vinod in making sure the test grid is all green.

If it at all not idiomatic or gets in the way after a bit, please revert.

PR #150, opened 2020-10-01
Release Google.Cloud.Functions.Framework version 1.0.0-alpha13

jskeet:
Changes since 1.0.0-alpha12:

- Modify the Hosting package MSBuild targets to only create an entry point for
  consuming projects with an output type of Exe.


amanda-tarafa (approved):
LGTM

PR #151, opened 2020-10-05
Fix autorelease build to use tags

jskeet:
This validates that the committish we build genuinely points at a
tag, and passes the tag to build-release (which expects one).

(This was basically copied from https://github.com/googleapis/google-cloudevents-dotnet/pull/48, but we already had --recursive in build-release.sh in this repo.)


amanda-tarafa (approved):
LGTM

PR #156, opened 2020-10-15
Add a test-oriented attribute for functions startups.

### LINKED ISSUE #154: Specify test startups on the test class as attributes
Current demo code:

```csharp
public class FunctionTest : FunctionTestBase<Function>
{
    public FunctionTest()
        : base(FunctionTestServerBuilder
                .Create<Function>()
                .UseFunctionsStartups(new FakeStartup())
                .Build())
    {
    }

    ...
}
```

This would be nicer:

```csharp
[FunctionsTestStartup(typeof(FunctionTest.FakeStartup))]
public class FunctionTest : FunctionTestBase<Function>
{
    ...
}
```

Note that it's not trying to reuse the existing `FunctionsStartupAttribute` because that can only be applied to the assembly - which is what we want. (At least, I think. We could potentially allow the function itself to have a startup attribute, but that's a different feature.)

jskeet:
Fixes #154.

An alternative approach would be to allow the existing
FunctionsStartupAttribute to be applied to types as well as the
assembly... in which case we'd probably want to change the framework
to find startup classes associated with the function itself. That
could potentially be quite useful...

Very happy to discuss those options.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -81,7 +81,7 @@ test frameworks that automatically dispose of test classes. (This
 includes NUnit and xUnit.) The `Dispose` implementation disposes of
 the test server that is owned by the test class.
 
-The parameterless constructor of `FunctionTestbase<TFunction>`
+The parameterless constructor of `FunctionTestBase<TFunction>`
 automatically creates a `FunctionTestServer<TFunction>`, or you can
 pass the test server into the constructor if you want to customize it.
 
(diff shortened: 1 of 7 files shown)


jskeet:
Looking at the test failure now.

jskeet:
I *tried* the "single attribute, which works on types" - but it doesn't work out.

The function target is determined by configuration - which is itself influenced by startup types. So there's a chicken and egg situation, basically. We'll go ahead with this PR.

amanda-tarafa (approved):
LGTM

PR #159, opened 2020-10-20
Revisit functions startups again

### LINKED ISSUE #158: Revisit applying startups via type attributes
While looking into #154, we investigated the possibility of having a single FunctionsStartupAttribute that could be applied to types as well as assemblies, so you could decorate the function.

That leads to a chicken and egg situation - startups can change the configuration, and the configuration can specify the function, which would in turn change which startups were applied.

But the frustrating thing is that in _almost all cases_ the function we expect to run before starting any complex processing is going to be the one we _actually_ run. So maybe we should go for a "trust and validate" model? Guess what the function target type is, include the startups from that function, and then if later on we determine that was incorrect (and that it would have changed which startups were invoked) we can throw an exception with a detailed explanation.

cc @amanda-tarafa - we'll want to discuss this.

jskeet:
This removes the test-specific attribute, allowing the same attribute to be used from both tests and production code, and for it to be specified at an assembly and type level.

There's a certain amount of fiddliness here which is only trying to address a very cornery corner case - but it's all in implementation rather than API.

None of this change is breaking compared with beta01, although it's breaking compared with "head" (so I'm glad we haven't released a beta02 yet).

Fixes #158


jskeet:
Oh, there's a very small breaking change in terms of UseFunctionsStartups now taking a type as well as an assembly, but that's all :)

src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
```diff
@@ -47,12 +48,12 @@ public FunctionTestServer(Type functionTarget) : this(functionTarget, null)
...
         /// <summary>
         /// Creates a FunctionTestServer for the specified function type, with an optional
-        /// type to inspect for <see cref="FunctionTestStartupAttribute"/> attributes.
+        /// type to inspect for <see cref="FunctionsStartupAttribute"/> attributes.
         /// </summary>
         /// <param name="functionTarget">The function type to host in the server.</param>
-        /// <param name="attributedStartupType">A type to examine for <see cref="FunctionTestStartupAttribute"/> attributes to provide further customization.
-        /// If this is null, or no attributes are specified in the type (or its base class hierarchy) the startup classes in the
-        /// function assembly are used.</param>
+        /// <param name="attributedStartupType">A type to examine for <see cref="FunctionsStartupAttribute"/> attributes to provide further customization.
```
amanda-tarafa, line 54: nit: The name of this param seems backwards to me. Maybe startupSourceType? typeWithStartups? maybe even startupAttributedType? As it is it just looks like we are passing the Startup type.

src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
```diff
@@ -47,12 +48,12 @@ public FunctionTestServer(Type functionTarget) : this(functionTarget, null)
...
+        /// If this is null, or no attributes are specified in the type (or its base class hierarchy) or the assembly it's declared in,
+        /// the startup classes in the function assembly are used.</param>
```
amanda-tarafa, line 56: Not just the ones in the function assembly right? Also the ones specified via attributes in the function itself? Maybe, "the function's usual startup classes will be used".

src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
```diff
@@ -127,11 +128,11 @@ public FunctionTestServer() : this(null)
 
         /// <summary>
         /// Constructs a new instance serving <typeparamref name="TFunction"/>, with an optional
-        /// type to inspect for <see cref="FunctionTestStartupAttribute"/> attributes.
+        /// type to inspect for <see cref="FunctionsStartupAttribute"/> attributes.
         /// </summary>
-        /// <param name="attributedStartupType">A type to examine for <see cref="FunctionTestStartupAttribute"/> attributes to provide further customization.
-        /// If this is null, or no attributes are specified in the type (or its base class hierarchy) the startup classes in the
-        /// function assembly are used.</param>
+        /// <param name="attributedStartupType">A type to examine for <see cref="FunctionsStartupAttribute"/> attributes to provide further customization.
```
amanda-tarafa, line 133: nit: same here with the name of the parameter, if you decide to change it.

src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
```diff
@@ -127,11 +128,11 @@ public FunctionTestServer() : this(null)
...
+        /// If this is null, or no attributes are specified in the type (or its base class hierarchy) or the assembly it's declared in,
+        /// the startup classes in the function assembly are used.</param>
```
amanda-tarafa, line 135: Same comment as before re "the usual startups..."

src/Google.Cloud.Functions.Testing/FunctionTestServerBuilder.cs
```diff
@@ -72,8 +72,9 @@ public FunctionTestServerBuilder UseFunctionsStartups(IEnumerable<FunctionsStart
         }
 
         /// <summary>
-        /// Examines <paramref name="attributedType"/> (and its base class hierarchy) for <see cref="FunctionTestStartupAttribute" />
-        /// annotations. If any such attributes are found, the startup classes specified by the attributes are used
+        /// Examines <paramref name="attributedType"/> (and its base class hierarchy, and the assembly in which it's declared)
+        /// for <see cref="FunctionsStartupAttribute" /> annotations.
+        /// If any such attributes are found, the startup classes specified by the attributes are used
         /// instead of the startup classes specified in the assembly containing the function target. If no attributes
```
amanda-tarafa, line 78: "instead of the usual startup classes for the target function..."

examples/Google.Cloud.Functions.Examples.IntegrationTests/TestableDependenciesTest.cs
```diff
@@ -23,12 +23,12 @@ namespace Google.Cloud.Functions.Examples.IntegrationTests
 {
     /// <summary>
     /// By default, the test server will use the same Functions Startup classes
-    /// as normal. The TestFunctionStartup attribute specifies Functions Startup classes that
-    /// should be used for this test class, so that it can inject test dependencies instead of the production ones.
+    /// as normal. Applying the FunctionsStartup attribute to the test class (or the assembly) signals to FunctionTestBase
```
amanda-tarafa, line 26: nit: (and/or the assembly)

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -84,6 +88,25 @@ internal static IApplicationBuilder ConfigureApplication(WebHostBuilderContext c
                 });
         }
 
+        /// <summary>
+        /// Validates that the startup classes which would be used for the finally-selected function are the
+        /// same as the startup classes we *actually* used. This method is only called when the app is started
```
amanda-tarafa, line 93: ...are the same and in the same order...

docs/customization.md
```diff
@@ -169,3 +199,40 @@ you genuinely intend to remove the Functions Framework console logging.
 
 Logging for tests is provided automatically in the `FunctionTestServer`; see [the testing
 documentation](testing.md) for more details.
+
+## Startup class mismatches
+
+If you haven't received an error referring you to this section of
+the documentation, you *may* want to skip this section.
+
+It is possible (but very rare) for startup classes to affect which
```
amanda-tarafa, line 208: Ahhh, there's a use case that could be until now, and now can't (a little far fetch though)

What if StartupA, say, depending on configuration, were to add another StartupB on it's own ConfigureServices? StartupB just overwrites the Configure method to add a few things to the pipeline later. That's sort of valid (and StartupB would have actually been executed properly) and sort of makes sense, but would fail now where it didn't before. Maybe we can allow for startup additions with the warning that, the ConfigureConfiguration override won't execute because it's already happened?

I'm actually torn between what should be invalid, a change in startups or a change in function type? Maybe later this can be configurable, like allowFunctionSwap or allowStartupAddition?

examples/Google.Cloud.Functions.Examples.AdvancedDependencyInjection/Function.cs
```diff
@@ -72,14 +65,25 @@ public class Startup : FunctionsStartup
...
     /// The actual Cloud Function using the dependencies.
     /// This uses both IOperationSingleton and IOperationScoped, and outputs the OperationId
     /// from each of them in HandleAsync. The service configuration provided in Startup
     /// means that for any given server, the SingletonOperationId property remains the same, but
     /// ScopedOperationId varies for each request.
     /// (Restarting the server will result in a different SingletonOperationId - nothing is persisted
     /// here.)
+    /// </para>
+    /// <para>
+    /// The FunctionsStartup attribute specifies the startup class to load. If you have multiple
```
amanda-tarafa, line 78: nit: `<see ref=FunctionsStartupAttribute />`?

amanda-tarafa (approved):
LGTM. A few nits, and a couple of comments/questions but I'm fine with this as is.

src/Google.Cloud.Functions.Hosting/FunctionsStartupAttribute.cs
```diff
@@ -43,5 +48,25 @@ public class FunctionsStartupAttribute : Attribute
...
+        /// <summary>
+        /// Returns the functions startup classes found in the attributes of <paramref name="assembly"/> and
+        /// <paramref name="target"/>, ordered by <see cref="Order"/> and then by the full name of the type.
+        /// </summary>
+        /// <param name="assembly">The assembly to inspect for attributes.</param>
+        /// <param name="target">Optional type to query for attributes. If this is null, only
+        /// startup classes from <paramref name="assembly"/> are returned.
+        /// </param>
+        /// <returns>A (possibly empty) sequence of the startup types detected via attributes.</returns>
+        public static IEnumerable<Type> GetStartupTypes(Assembly assembly, Type? target)
```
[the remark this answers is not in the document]
jskeet: Yes, the assembly of the `target` should be the same as `assembly`. (If `target` were required, I'd drop `assembly` entirely.) At least, I can't think of any times when they'd be different for production code. For test code it's just about plausible, although we wouldn't use that ourselves.

Do you think it's worth validating that? We could always relax it later.

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -156,14 +157,14 @@ internal static IServiceCollection AddServicesForFunctionTarget(IServiceCollecti
         /// discover problems there easily enough anyway.)
         /// </summary>
         /// <remarks>
-        /// This method is internal and in this class rather than being private and in Builder for the sake of testability.
+        /// This method is internal rather private for the sake of testability.
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -172,7 +173,33 @@ internal static Type FindDefaultFunctionType(params Type[] types)
...
 
-            bool IsFunctionClass(Type t) =>
+        }
+
+        /// <summary>
+        /// Attempts to find a single valid non-abstract function class within the given set of types, but returns
+        /// null (instead of failing) if there isn't exactly one such class.
+        /// </summary>
+        /// <param name="types"></param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
jskeet: Done.

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -113,15 +105,24 @@ internal static async Task Execute(HttpContext context)
...
                 : assembly.GetType(target) ?? throw new InvalidOperationException($"Can't load specified function type '{target}'");
         }
 
+        internal static Type? TryGetFunctionTarget(IConfiguration configuration, Assembly assembly)
+        {
+            var options = FunctionsFrameworkOptions.FromConfiguration(configuration);
+            string? target = options.FunctionTarget;
+            return target is null
+                ? TryFindDefaultFunctionType(assembly.GetTypes())
+                : assembly.GetType(target) ?? throw new InvalidOperationException($"Can't load specified function type '{target}'");
```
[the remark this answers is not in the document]
jskeet: Hmm. I guess it's just about conceivable that we get given a type name that's invalid, but then a startup class replaces it with a valid one... Yup, removed.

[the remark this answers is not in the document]
jskeet: (In particular, if we're *not* in that weird situation, we'll still fail clearly later.)

src/Google.Cloud.Functions.Hosting/EntryPoint.cs
```diff
@@ -64,13 +70,21 @@ public static async Task<int> StartAsync(Assembly functionAssembly, string[] arg
...
             await host.RunAsync();
             return 0;
+
+            Type? GuessFunctionTarget()
+            {
+                var configuration = new ConfigurationBuilder()
+                    .AddFunctionsEnvironment()
+                    .AddFunctionsCommandLine(args)
+                    .Build();
+                return HostingInternals.TryGetFunctionTarget(configuration, functionAssembly);
```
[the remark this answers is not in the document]
jskeet: Yup, concur.

src/Google.Cloud.Functions.Testing/FunctionTestServerBuilder.cs
jskeet: Done.

examples/Google.Cloud.Functions.Examples.IntegrationTests/TestableDependenciesTest.cs
jskeet: Done.

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
jskeet: Done.

docs/customization.md
jskeet: Adding a startup is really dodgy, because it would only be used for middleware, not config or dependencies. I don't think that's something we want to support, and any previous support is accidental.

We can make this sort of check optional later on, maybe... but I think it's massively corner-case territory.

examples/Google.Cloud.Functions.Examples.AdvancedDependencyInjection/Function.cs
jskeet: I've deliberately not done that because it makes it harder to read just in GitHub, which is where I'm expecting it to be read more than in VS. A VS user can always click through via the attribute itself if necessary.

PR #160, opened 2020-10-20
Multifunction example

jskeet:
Only the last commit is worth looking at - the rest is from #159.


amanda-tarafa (commented):
Yep, I agree this a little too much and might be more confusing than useful.

jskeet:
Great - thanks for the review. Closing.

PR #164, opened 2020-10-23
Minor F# changes

isaacabraham:
Some minor changes to the three F# samples. Hopefully they're useful.

1. The previous use of `async` was semantically incorrect / different from the C# Task version. The current F# example creates the task but never "awaits" it, it just throws it away (with `ignore`) - this is the same as in C# simply not awaiting an async (whereby you get the compiler warning). `do!` is the equivalent to `await` in this context, so I've put that in here.
2. Replace use of %A (which is a kind of magic "print out this value") with either specific modifiers for longs (%i) or ToString (%O).
3. Added the "Time" item which was in the C# examples. F# is a bit (ok, a lot) more verbose here because:
   a. There's no language support for mapping / binding on nullable values, so instead you use `Option.map`.
   b. There's no implicit conversions so we need `Option.ofNullable` to get "into" the Option world.
   c. There's no implicit "default" value, so we explicitly give a default of "" using `Option.defaultValue`

### DIFF
--- a/examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
+++ b/examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
@@ -40,14 +40,18 @@ type Function() =
             printfn "Storage object information:"
             printfn "  Name: %s" data.Name
             printfn "  Bucket: %s" data.Bucket
-            printfn "  Size: %A" data.Size
+            printfn "  Size: %i" data.Size
             printfn "  Content type: %s" data.ContentType
             printfn "Cloud event information:"
             printfn "  ID: %s" cloudEvent.Id
-            printfn "  Source: %A" cloudEvent.Source
+            printfn "  Source: %O" cloudEvent.Source
             printfn "  Type: %s" cloudEvent.Type
             printfn "  Subject: %s" cloudEvent.Subject
-            printfn "  DataSchema: %A" cloudEvent.DataSchema
-            printfn "  DataContentType: %A" cloudEvent.DataContentType
-            printfn "  SpecVersion: %A" cloudEvent.SpecVersion
+            printfn "  DataSchema: %O" cloudEvent.DataSchema
+            printfn "  DataContentType: %O" cloudEvent.DataContentType
+            printfn "  Time: %s" (cloudEvent.Time |> Option.ofNullable |> Option.map(fun time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'") |> Option.defaultValue "")
+            printfn "  SpecVersion: %O" cloudEvent.SpecVersion
+
+            // In this example, we don't need to perform any asynchronous operations, so we
+            // just return an arbitrary Task to conform to the interface.
             Task.CompletedTask
--- a/examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
+++ b/examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
@@ -49,9 +49,11 @@ type Function() =
             printfn "  Subject: %s" cloudEvent.Subject
             printfn "  DataSchema: %O" cloudEvent.DataSchema
             printfn "  DataContentType: %O" cloudEvent.DataContentType
-            printfn "  Time: %s" (cloudEvent.Time |> Option.ofNullable |> Option.map(fun time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'") |> Option.defaultValue "")
+            printfn "  Time: %s" (match Option.ofNullable cloudEvent.Time with
+                                  | Some time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'"
+                                  | None -> "")
             printfn "  SpecVersion: %O" cloudEvent.SpecVersion
 
             // In this example, we don't need to perform any asynchronous operations, so we
-            // just return an arbitrary Task to conform to the interface.
+            // just return an completed Task to conform to the interface.
             Task.CompletedTask
--- a/examples/Google.Cloud.Functions.Examples.FSharpHttpFunction/Function.fs
+++ b/examples/Google.Cloud.Functions.Examples.FSharpHttpFunction/Function.fs
@@ -26,5 +26,5 @@ type Function() =
         /// <returns>A task representing the asynchronous operation.</returns>
         member this.HandleAsync context =
             async {
-                context.Response.WriteAsync "Hello, Functions Framework." |> Async.AwaitTask |> ignore
+                do! context.Response.WriteAsync "Hello, Functions Framework." |> Async.AwaitTask
             } |> Async.StartAsTask :> _
--- a/examples/Google.Cloud.Functions.Examples.FSharpUntypedEventFunction/Function.fs
+++ b/examples/Google.Cloud.Functions.Examples.FSharpUntypedEventFunction/Function.fs
@@ -39,5 +39,5 @@ type Function() =
             printfn "Data: %A" cloudEvent.Data
 
             // In this example, we don't need to perform any asynchronous operations, so we
-            // just return an arbitrary Task to conform to the interface.
+            // just return a completed Task to conform to the interface.
             Task.CompletedTask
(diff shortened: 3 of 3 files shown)


jskeet:
Thanks very much for this - I'll look in detail on Monday

examples/Google.Cloud.Functions.Examples.FSharpUntypedEventFunction/Function.fs
```diff
@@ -33,7 +33,11 @@ type Function() =
...
-            printfn "DataContentType: %A" cloudEvent.DataContentType
-            printfn "SpecVersion: %A" cloudEvent.SpecVersion
+            printfn "DataContentType: %O" cloudEvent.DataContentType
+            printfn "Time: %s" (cloudEvent.Time |> Option.ofNullable |> Option.map(fun time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'") |> Option.defaultValue "")
+            printfn "SpecVersion: %O" cloudEvent.SpecVersion
             printfn "Data: %A" cloudEvent.Data
-            Task.CompletedTask
+
+            // In this example, we don't need to perform any asynchronous operations, so we
+            // just return an arbitrary Task to conform to the interface.
```
jskeet, line 42: (Here and earlier.) It's not entirely arbitrary - it's important that it's completed.
```suggestion
            // just return a completed Task to conform to the interface.
```

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
```diff
@@ -40,14 +40,18 @@ type Function() =
...
+            printfn "  Source: %O" cloudEvent.Source
             printfn "  Type: %s" cloudEvent.Type
             printfn "  Subject: %s" cloudEvent.Subject
-            printfn "  DataSchema: %A" cloudEvent.DataSchema
-            printfn "  DataContentType: %A" cloudEvent.DataContentType
-            printfn "  SpecVersion: %A" cloudEvent.SpecVersion
-            Task.CompletedTask
+            printfn "  DataSchema: %O" cloudEvent.DataSchema
+            printfn "  DataContentType: %O" cloudEvent.DataContentType
+            printfn "  Time: %s" (cloudEvent.Time |> Option.ofNullable |> Option.map(fun time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'") |> Option.defaultValue "")
```
jskeet, line 52: This definitely brings it into parity with the C# example... but do you think it's worth the extra complexity? I'm concerned that the impedance mismatch here ends up looking like the most important aspect of the function, given the amount of code dedicated to it.

examples/Google.Cloud.Functions.Examples.FSharpHttpFunction/Function.fs
```diff
@@ -26,5 +26,5 @@ type Function() =
         /// <returns>A task representing the asynchronous operation.</returns>
         member this.HandleAsync context =
             async {
-                context.Response.WriteAsync "Hello, Functions Framework." |> Async.AwaitTask |> ignore
+                do! context.Response.WriteAsync "Hello, Functions Framework." |> Async.AwaitTask
```
jskeet, line 29: Thanks! Much appreciated :) (I did experiment for a while trying to get this to work, but must have missed the option of `do!`.)

jskeet:
Ping for the requested changes and CLA? I'm hoping to do a new release soon, and it would be great to get the changes in.

jskeet:
(CLA is now green, so it's really just the suggestion about the task comment, and whether or not it's worth the complexity of displaying the timestamp.)

isaacabraham:
@jskeet apologies, just seen your comments. I'll do them this week.

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
isaacabraham: Mmm, I had the same thought. Let me look at it again to see :-)

jskeet:
The CLA issue is now due to a joint commit with my personal account :) Will sort that out.

jskeet:
Added "cla: yes" label due to the suggestions being accepted with my personal email address. (I'd expect the CLA bot to find that anyway, but let's not worry right now.)

Reviewing again...

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
```diff
@@ -40,14 +40,18 @@ type Function() =
...
+            printfn "  SpecVersion: %O" cloudEvent.SpecVersion
+
+            // In this example, we don't need to perform any asynchronous operations, so we
+            // just return an arbitrary Task to conform to the interface.
```
jskeet, line 56: Another "arbitrary" task to be changed to "completed" as per the untyped event function.

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
jskeet: Any more thoughts on this? We're nearly there :)

isaacabraham:
Sorry - I haven't done this yet :-( I think it's worth me seeing if I can make it a bit cleaner, maybe using pattern matching rather than the map / defaultValue stuff. If that doesn't work, perfectly happy for you to remove those fields in the interest of clarity.

jskeet:
Cool - I'm very happy to wait for you to have another look, and thank you again for taking the time in the first place!

isaacabraham:
Quick question - what's the plan vis-a-vis .NET 5 (and F#5)? F#5 has support for string interpolation, which is both something C# developers are aware of, and more succinct than `sprintf`.

isaacabraham:
Then again as everything here is `printfn` (and not `sprintf`) it might not be worthwhile.

jskeet:
I'm hoping .NET 5 will eventually be supported as a runtime, but as .NET Core 3.1 is the LTS, that will be our "default" platform. So I don't want to add dependencies on F# 5 at the moment, I'm afraid - much as I can see it being cleaner.

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
isaacabraham: Fixed.

examples/Google.Cloud.Functions.Examples.FSharpEventFunction/Function.fs
isaacabraham: I've changed this into a pattern match over three lines. I think it's more readable. However, if you would prefer, let me know and I'll just remove the entire Time line in both files.

examples/Google.Cloud.Functions.Examples.FSharpUntypedEventFunction/Function.fs
```diff
@@ -33,7 +33,15 @@ type Function() =
...
+                                | Some time -> time.ToUniversalTime().ToString "yyyy-MM-dd'T'HH:mm:ss.fff'Z'"
+                                | None -> "")
+            printfn "SpecVersion: %O" cloudEvent.SpecVersion
             printfn "Data: %A" cloudEvent.Data
+
+            // In this example, we don't need to perform any asynchronous operations, so we
+            // just return a completed Task to conform to the interface.
             Task.CompletedTask
+
+let x = Nullable 10
```
jskeet, line 47: Was this intentional? My guess is you were checking something else...

jskeet (commented):
Yup, that looks fine to me. Just the stray error at the end of the untyped function to remove, then we can merge.

If we can get this done and released today, that would be really good...

examples/Google.Cloud.Functions.Examples.FSharpUntypedEventFunction/Function.fs
isaacabraham: Doh!!! That was my just testing out if you could pattern match on Nullables directly (you can't).

jskeet: Just checking - are you pushing a new commit to remove this? I think we're ready to release when you do.

isaacabraham: Yep, I am.

isaacabraham: Pushed.

jskeet:
Thanks! Will squash and merge on green.
(Again, the CLA part is due to the commit with my personal email involved. Overriding it is fine.)

The conformance test failure is also okay - that's a known issue with the conformance tests, basically.

jskeet (approved):

PR #166, opened 2020-11-03
Fix xml comment compiler warning

alteredego:


jskeet:
Will just have a look before we merge this as to why I didn't spot it before...

jskeet:
Hmm... I can't see a warning here at all, and in VS I can use F12 to get to `LogImpl<TState>`. I don't object to the change, but could you let me know what's warning you? Is this a Resharper warning, for example?

alteredego:
You're absolutely right, it must be R# giving the warning - "Cannot resolve reference in XML comment".  It's the only error it shows in the solution and underlines the reference, but F12 does jump to the correct place in Visual Studio without the change.  I think the suggested change is "correct" though.

jskeet:
Righto - that's fine then; will merge. I just wanted to check that I didn't need to change any warning flags etc first.

jskeet (approved):

PR #177, opened 2020-12-07
Add Firebase Analytics custom conversion and sample data

jskeet:
(This should be the last event type we need to support, as far as I'm aware.)


amanda-tarafa (approved):
LGTM

PR #178, opened 2020-12-08
Prototype of what a dynamic function might look like

jskeet:
This is definitely not ready to be submitted, but is presented as a "what if" thought...

Thoughts very welcome.

### DIFF
--- a/src/Google.Cloud.Functions.Framework/DynamicFunctionAdapter.cs
+++ b/src/Google.Cloud.Functions.Framework/DynamicFunctionAdapter.cs
@@ -1,0 +1,80 @@
+﻿// Copyright 2020, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Framework
+{
+    /// <summary>
+    /// Function which delegates to an IDynamicFunction.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


src/Google.Cloud.Functions.Framework/DynamicFunctionAdapter.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Framework
+{
+    /// <summary>
+    /// Function which delegates to an IDynamicFunction.
+    /// </summary>
+    public sealed class DynamicFunctionAdapter : IHttpFunction
+    {
+        // TODO: *Lots* of caching by type. We may not be able to avoid reflection for the final invocation,
```
amanda-tarafa, line 30: Just a thought that might be off the mark. What if on startup, when we detect a dynamic function, we emit a class with a "normal function" that has a constructor with the parameters from the dynamic `HandleAsync` and wraps the dynamic function, and we use that type as the startup function, so that the activator does the dependency injection bits. Still we'd have to use reflection to make the final call as well as here, but we don't have to deal with DI directly. I think.

amanda-tarafa (commented):
Yes, I like this.
I've left a couple of comments, and apart from that, the name Dynamic doesn't inmediately bring this to mind, but I don't have a better name.

src/Google.Cloud.Functions.Framework/DynamicFunctionAdapter.cs
```diff
@@ -0,0 +1,80 @@
...
+            (_function, _serviceProvider) = (function, serviceProvider);
+        }
+
+        /// <summary>
+        /// Implements the IHttpFunction contract by executing the Handle or HandleAsync method dynamically.
+        /// </summary>
+        public Task HandleAsync(HttpContext context)
+        {
+            // TODO: Adapt sync functions too.
+            var method = _function.GetType().GetMethod("HandleAsync", BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
```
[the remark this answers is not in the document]
jskeet: Just thought it might be nice - but yes, it does prevent overloads.

src/Google.Cloud.Functions.Framework/DynamicFunctionAdapter.cs
jskeet: If we actually emit a class, and if we only use public methods, we don't need to use any reflection for the invocation - we can just generate code that calls the method.

The code would be pretty hairy, but it would certainly fly... it's an intriguing idea.

amanda-tarafa: Ah yes, restricting to public methods has that advantage.

jskeet:
Closing as I don't think we'll want to do this any time soon, although I'll leave the branch available.

PR #183, opened 2021-01-12
Dispose of the host instead of the TestServer in FunctionTestServer

### LINKED ISSUE #182: TestServer process hanging around after test
In Rider after I run a test that inherits from `FunctionTestBase` I get the below warning from Rider after a while.  I have usually seen this when `TestServer` or the `HttpClient` from calling `TestServer.CreateClient` is not disposed but after a brief look into` FunctionTestBase` and `FunctionTestServer` the `Dispose` methods look ok and clients are marked inside a using clause so I'm unsure what causes this warning. 

Perhaps my gut feeling is wrong and something else is keeping the process alive. 

I'm on MacOS Big Sur.

Screenshot 2021-01-08 at 11 30 46

jskeet:
Fixes #182.

It feels like this is really a bug in TestServer, but it's easy
enough to fix it in FunctionTestServer without introducing a
breaking change.

Given that TestServer.Host throws, I've made the IHost available in
FunctionTestServer too.

### DIFF
--- a/src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
+++ b/src/Google.Cloud.Functions.Testing/FunctionTestServer.cs
@@ -15,6 +15,7 @@
 using Google.Cloud.Functions.Hosting;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


amanda-tarafa (approved):
LGTM

PR #191, opened 2021-01-21
Add (and use) ConfigureFunctionsFramework extension method

### LINKED ISSUE #181: Accessing UseServiceProviderFactory method
Hi,

Is it possible to call UseServiceProviderFactory somehow? In ordinary projects, We are doing it in Main function in Program.cs. But now it looks we don't have access to Program.cs.

jskeet:
This allows for user customization of the host builder before we get to it (via a custom entry point)

Fixes #181

### DIFF
--- a/src/Google.Cloud.Functions.Hosting/EntryPoint.cs
+++ b/src/Google.Cloud.Functions.Hosting/EntryPoint.cs
@@ -13,10 +13,7 @@
 // limitations under the License.
 
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
-using Microsoft.Extensions.Logging;
 using System;
 using System.Reflection;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


jskeet:
(Example needed, when I've confirmed that this is what we want.)

gkinsman:
Hi there! In case you're after another example simpler than Autofac integration, may I suggest using Serilog as the root logger instead of MSEL? 

The integration point is [here](https://github.com/serilog/serilog-aspnetcore/blob/dev/src/Serilog.AspNetCore/SerilogWebHostBuilderExtensions.cs), and adds a few things to services too.

This would fit your PR scenario well I think, as with Serilog its a good idea to wrap the entry point in a try/finally to catch startup errors, and to configure the logger before the app starts.

jskeet:
@gkinsman: I'd happily revisit this PR if it becomes necessary, but this was specifically requested for Autofac integration.

Can Serilog integration not be done just with ConfigureLogging? If not, and if you're an interested customer requesting Serilog integration, I can look at it - but I'd rather not spend the time going back to this unless it's really required.

jskeet:
Closing as I don't think we'll want to do this any time soon, although I'll leave the branch available.

PR #195, opened 2021-01-26
docs: Deploying an example with a modified Functions Framework

jskeet:


examples/README.md
```diff
@@ -34,3 +34,20 @@ the `src` directory), set an MSBuild property of
...
 environment variable set.
+
+To deploy an example using the local version of the Functions
+Framework, you can use `--set-build-env-vars` to set the
+`LOCALFUNCTIONSFRAMEWORK` environment variable (which must be in
+all-caps for the Buildpack to work) to true, along with
+`GOOGLE_BUILDABLE` to select the project. This has to be run from
+the root directory, in order to pick up `src` as well as `examples`.
+So for example, to run the `SimpleHttpFunction` example using a
+modified Functions Framework, you could run:
```
amanda-tarafa, line 45: nit: Is the use of `src` and `examples`  here very particular of running with the same folder structure as we have. If they have modified the framework, chances are they have moved it around? Not sure how to rephrase, but maybe move the mention of `src` and `examples` to the "So for example" sentence.

amanda-tarafa (approved):
LGTM

examples/README.md
jskeet: This file is *in* the examples directory... the paragraph is specifically about wanting to run these examples. I'd expect anyone wanting to tweak the framework to basically clone the whole repo, and the simplest way of prototype-testing a change is via the examples here. Expanding these instructions to include scenarios where the framework and the function are miles apart would make them much more complex.

Will leave for now - we can revisit if necessary. (This whole change is probably more for my benefit than anyone else's :)

amanda-tarafa: SGTM

PR #197, opened 2021-02-10
Update handling of Firebase Auth events

jskeet:


jskeet:
We'll need to update the conformance test before this passes... I'll do that tomorrow, then update this PR with another commit that moves the submodule. I don't expect any further *code* changes though, so it can be reviewed while red if that's convenient.

amanda-tarafa (approved):
LGTM, but conformance tests need updating as well I think.

PR #213, opened 2021-06-17
Support the location part of the Firebase RTDB source

jskeet:
With this in place, the source and subject for all events are correct as per internal documentation.

This PR will fail conformance tests until https://github.com/GoogleCloudPlatform/functions-framework-conformance/pull/76 has been merged, at which point I'll add a commit to update our submodule to that. (The two PRs pass together locally.)


amanda-tarafa (approved):
LGTM

PR #216, opened 2021-06-22
Remove version number from template installation instructions

jskeet:
(It's not needed now that the package has hit 1.0.0.)


jskeet:
(Will self-merge, as it's so trivial.)

PR #217, opened 2021-07-08
Update example dependencies

jskeet:
Most importantly, this updates Google.Cloud.Vision.V1 to a version which transitively depends on a much smaller version of Grpc.Core


amanda-tarafa (approved):
LGTM

PR #218, opened 2021-07-16
Update CloudEvents SDK dependencies

jskeet:
The code changes are to handle the nullable reference types which
are now in the SDK. The Functions Framework itself still doesn't
advertise anything around nullable reference types - that's a change
that *could* come later if we want to.


amanda-tarafa (approved):
LGTM

jskeet:
Will update commit to follow conventional commits...

PR #230, opened 2021-11-24
First commit of extra examples

jskeet:
This PR is to investigate possible use cases, and how easily they could be achieved.

### DIFF
--- a/extra-examples/CustomPayload2/Function.cs
+++ b/extra-examples/CustomPayload2/Function.cs
@@ -1,0 +1,45 @@
+// Copyright 2021, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using CloudNative.CloudEvents;
+using CloudNative.CloudEvents.NewtonsoftJson;
+using Google.Cloud.Functions.Framework;
+using Google.Cloud.Functions.Hosting;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using ThirdPartyPackage;
+
+namespace CustomPayload2
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
+            services.AddSingleton<CloudEventFormatter, JsonEventFormatter<Payload>>();
+    }
+
+    [FunctionsStartup(typeof(Startup))]
+    public class Function : ICloudEventFunction<Payload>
+    {
+        public Task HandleAsync(CloudEvent cloudEvent, Payload data, CancellationToken cancellationToken)
+        {
+            Console.WriteLine($"ID: {cloudEvent.Id}");
+            Console.WriteLine($"Text: {data.Text}");
+            Console.WriteLine($"Number: {data.Number}");
+            return Task.CompletedTask;
+        }
+    }
+}
--- a/extra-examples/GoogleTypeDetection/Function.cs
+++ b/extra-examples/GoogleTypeDetection/Function.cs
@@ -1,0 +1,55 @@
+// Copyright 2021, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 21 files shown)


extra-examples/CustomPayload2/Function.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using ThirdPartyPackage;
+
+namespace CustomPayload2
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
```
matthewrobertson, line 30: It is a bit odd to me that the formatter is configured as a Functions Framework service and not on the Google Events library. Said another way: how do we expect customers will use the Google Events library outside of the Functions Framework? For example consider the use case of deploying an ASP.NET app to CloudRun and point an EventArc subscription at it. I would expect to be able to do something like:


```dotnet
// using stuff
var builder = WebApplication.CreateBuilder(args);

// I guess would need to configure a service here?

var app = builder.Build();

app.MapPost("/", (CloudEvent cloudEvent) => { 
  // now I want to work with a Google Event data payload...
});
```

extra-examples/GoogleTypeDetection/Function.cs
```diff
@@ -0,0 +1,55 @@
...
+                case StorageObjectData storage:
+                    Console.WriteLine($"Storage Bucket: {storage.Bucket}");
+                    break;
+                case MessagePublishedData pubsub:
+                    Console.WriteLine($"PubSub subscription: {pubsub.Subscription}");
+                    break;
+                default:
+                    Console.WriteLine($"Unhandled event type {cloudEvent.Type}");
+                    break;
+            }
```
matthewrobertson, line 51: 😍

extra-examples/GoogleTypeDetection/GoogleEventFormatter.cs
```diff
@@ -0,0 +1,138 @@
...
+            { DocumentEventData.CreatedCloudEventType, DocumentEventData.Descriptor },
+            { DocumentEventData.DeletedCloudEventType, DocumentEventData.Descriptor },
+
+            { ReferenceEventData.CreatedCloudEventType, ReferenceEventData.Descriptor },
+            { ReferenceEventData.DeletedCloudEventType, ReferenceEventData.Descriptor },
+            { ReferenceEventData.WrittenCloudEventType, ReferenceEventData.Descriptor },
+            { ReferenceEventData.UpdatedCloudEventType, ReferenceEventData.Descriptor },
+
+            // Etc... this would all be generated.
+        };
```
matthewrobertson, line 57: This is looks very close to what I had in mind for Typescript.

extra-examples/GoogleTypeDetection/Function.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GoogleTypeDetection
+{
+    public class Startup : FunctionsStartup
+    {
+        public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services) =>
+            services.AddSingleton<CloudEventFormatter, GoogleEventFormatter>();
```
matthewrobertson, line 31: Open question: can we get to a point where this is the pre-configured default in the Functions Framework so developers don't need to provide a `FunctionsStartup` to for this use case.

extra-examples/CustomPayload2/Function.cs
jskeet: This is in a scenario where you're not using the Google Events library at all - it's not a Google event, it's a custom event payload.

Using the Google Events library outside the Functions Framework, you'd need to set up code to receive the CloudEvent - that's one area of the SDK that's still slightly TBD, but doesn't feel like it's in the remit of the Google Events library particularly - that should make it easy to parse the payload as a Google event payload, but that's all. (If you're writing a web server to handle a non-Google-payload CloudEvent request, it's not really our business, even if it happens that the request is made by Eventarc.)

extra-examples/GoogleTypeDetection/Function.cs
```diff
@@ -0,0 +1,55 @@
...
+            Console.WriteLine($"ID: {cloudEvent.Id}");
+            switch (cloudEvent.Data)
+            {
+                case StorageObjectData storage:
+                    Console.WriteLine($"Storage Bucket: {storage.Bucket}");
+                    break;
+                case MessagePublishedData pubsub:
+                    Console.WriteLine($"PubSub subscription: {pubsub.Subscription}");
+                    break;
+                default:
```
matthewrobertson, line 48: it is possible to combine this with the 3rd party sample somehow so I can have a case like:

```csharp
case MyServiceSpecificPayload payload:
  break;
```

extra-examples/GoogleTypeDetection/Function.cs
jskeet: See the readme for more thoughts on this. We could potentially have an `IGoogleCloudEvent` marker interface that's implemented by all the concrete types, and then decorate that with the formatter reference. Customers would then write an `ICloudEventFunction<IGoogleCloudEvent>` implementation.

Having this as a vanilla default is also easy if we're happy to a) make the Functions Framework depend on the Google Events library; b) favour Google event payloads over other payloads. Lots of options to talk about...

extra-examples/GoogleTypeDetection/Function.cs
jskeet: I think that gets tricky - somewhere you've got to know where to get all the event types from. But we can think about it...

matthewrobertson: I think this is probably less common / lower priority use case so I don't think we need to go out of our way to optimize for it.

PR #235, opened 2022-01-04
docs: Demo of a PubSub "proxy" function for local development.

jskeet:
This is an HTTP function which is configured to make requests to a
CloudEvent PubSub function. It is designed to receive requests from
the PubSub emulator (via a subscription with a push config),
allowing for local development where normally an application would
publish PubSub messages and a CloudEvent function would consume them.


jskeet:
This is no longer needed, due to #238.

PR #237, opened 2022-01-25
Example of 'Reading' from Firestore

Equalateral:
NOTE: if I am mistaken the Google.Cloud.Firestore dependency needs to be installed in the local machine so that it is available in the deployment.  I have provided instruction in the attached as to how I configured the build.  

First MANY thanks for your efforts, I do not mean my comments or observations to come across as negative, I am just trying to be succinct.

You may feel free to remove what instruction you feel is redundant, as I say there are a LOT of ways to get this wrong, and mostly the deployment fails silently or there is just no response.  This had be on the verge of reverting back to Azure, just to have more mature dotnet workflows.  A big part of the problem (took me to the breaking point and I really cannot understand what exactly is going on is the "Type Tuples" [FirestoreData] & [FirestoreProperty] They aren't types and they aren't tuples but they APPEAR to be Required for the "ConvertTo()" to correctly destructure the DocumentSnapshot.  

 Please be aware that there are MANY ways for this to fail and when testing with a deployed function you will get precious little feedback
    If you are having trouble; pleas check that... 
    1)    The Firestore DB must be defined and populated CORRECTLY.
    2)    The deploy is correctly configured (verify the project-ids between Firestore DB and Function are aligned.
    3)    The payload in the function test aligns with the structure of the Firestore document

### DIFF
--- a/examples/Google.Cloud.Functions.Examples.FirestoreRead/Function
+++ b/examples/Google.Cloud.Functions.Examples.FirestoreRead/Function
@@ -1,0 +1,82 @@
+// Copyright 2022, Equalation LTD Co.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//  This example requires installation of Google.Cloud.Firestore from Nuget.
+//    Below is using Google.Cloud.Firestore v2.5.0
+//   Author is using Visual Studio 2022, WITH "Cloud Tools for Visual Studio"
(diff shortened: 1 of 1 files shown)


jskeet:
I don't think it's really worth having an example specifically for Firestore, to be honest... particularly when it seems that most of the issues have had are really around the Google.Cloud.Firestore library itself rather than the function. It's not entirely clear to me what the use case you're trying to solve is. (Yes, when you parse an HTTP request body as JSON, it needs to be valid JSON with the expected structure, but that doesn't seem unusual... likewise that you have to get the right project ID when integrating with Firestore.)

Rather than creating this as an example, I suggest you file individual, specific issues for aspects you found unexpected and where you think we should improve things. I suspect that the fact that you're using two new things (the Functions Framework and the Firestore library) at the same time has made it harder to tease those apart. I'd personally suggest experimenting with Firestore just using a regular console application, then experimenting with the Functions Framework in a standalone function that just logs - once those aspects are both working individually *then* it makes sense to put them together.

jskeet:
Additionally, I would note that for this:

> Please be aware that there are MANY ways for this to fail and when testing with a deployed function you will get precious little feedback

... any failure at execution time should be logged in the function in a pretty readable way. If you're not seeing a log entry for the exception, that's definitely an issue.

Equalateral:
Thanks for the consideration and feedback.  Please feel free to 

Steve

PR #238, opened 2022-02-15
feat: Support Pub/Sub push notifications format (adapt to CloudEvent)

### LINKED ISSUE #234: Fix local development with Pub/Sub Emulator
Hi there,

Currently the dotnet functions framework doesn't seem to work with the pubsub emulator, due to I believe the same issue as with other functions framework implementations (https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/272).

I've just hit this, as I'm receiving an error that isn't occurring when the function runs against real pub/sub.

```
[Google.Cloud.Functions.Framework.CloudEventAdapter] [fail] Event is malformed; does not contain a payload, or the event ID is missing.
```

jskeet:
Fixes #234

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -95,6 +95,29 @@ Firestore event          | DocumentEventData     | --trigger-event providers/clo
 >   available in the `FirestoreEvent` class via the `Wildcards` property. This is subject to change,
 >   as it's inconsitent with other Functions Frameworks.
 
+### Triggering an HTTP function with a Cloud Pub/Sub push subscription
+
+HTTP functions can work as the endpoints for [Cloud Pub/Sub push
+subscriptions](https://cloud.google.com/pubsub/docs/push). If your
+function implements `ICloudEventFunction` or
+`ICloudEventFunction<MessagePublishedData>`, the Functions Framework
+will adapt the incoming HTTP request to present it to your function
+as a CloudEvent, as if you had deployed via `--trigger-topic`. The
+requests for push subscriptions do not contain topic data, but if
+you create the push subcription with a URL of
+`https://<your-function>/projects/<project-id>/topics/<topic-id>`
+then the Functions Framework will infer the topic name from the path
+of the HTTP request. If the topic name cannot be inferred
+automatically, a topic name of
+`projects/unknown-project!/topics/unknown-topic!` will be used in
+the CloudEvent.
+
+> Note: the Functions Framework code to adapt the incoming HTTP
+> request works with the [Cloud Pub/Sub
+> Emulator](https://cloud.google.com/pubsub/docs/emulator), but
+> some information is not included in emulator push subscription
+> requests.
+
 ### Deploying a function with a local project dependency
 
 Real world functions are often part of a larger application which
--- a/src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
+++ b/src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
@@ -75,6 +75,7 @@ namespace Google.Cloud.Functions.Framework.GcfEvents
             internal const string StorageObjectChanged = StorageObjectV1 + ".changed";
         }
 
+        internal const string DefaultRawPubSubTopic = "projects/unknown-project!/topics/unknown-topic!";
         private const string JsonContentType = "application/json";
 
         private static readonly Dictionary<string, EventAdapter> s_eventTypeMapping = new Dictionary<string, EventAdapter>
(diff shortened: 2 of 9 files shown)


jskeet:
Note: tested with both the PubSub emulator and the production service

src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
```diff
@@ -292,6 +294,63 @@ protected override void MaybeReshapeData(Request request)
...
+                }
+                if (request.Data is object ||
+                    request.Domain is object ||
+                    request.Context is object ||
+                    request.EventId is object ||
+                    request.EventType is object ||
+                    request.Params is object ||
+                    request.Timestamp is object)
+                {
+                    throw new ConversionException("Request is malformed; raw Pub/Sub request must object contain 'message' and 'subscription' properties.");
```
amanda-tarafa, line 331: nit:
```suggestion
                    throw new ConversionException("Request is malformed; raw Pub/Sub request must object contain 'message' and 'subscription' properties only.");
```

amanda-tarafa (approved):
LGTM

src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
jskeet: Thanks - I'll probably put the "only" before the property names rather than after, but it's good to have that.

jskeet: Done (and force-pushed).

src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
```diff
@@ -292,6 +294,63 @@ protected override void MaybeReshapeData(Request request)
...
+                }
+                if (request.Data is object ||
+                    request.Domain is object ||
+                    request.Context is object ||
+                    request.EventId is object ||
+                    request.EventType is object ||
+                    request.Params is object ||
+                    request.Timestamp is object)
+                {
+                    throw new ConversionException("Request is malformed; raw Pub/Sub request must object contain only 'message' and 'subscription' properties.");
```
amanda-tarafa, line 331: actually (sorry for not catching earlier):
```suggestion
                    throw new ConversionException("Request is malformed; raw Pub/Sub request must contain only 'message' and 'subscription' properties.");
```

amanda-tarafa (approved):

src/Google.Cloud.Functions.Framework/GcfEvents/GcfConverters.cs
jskeet: Whoops, thanks, fixed.

docs/deployment.md
```diff
@@ -95,6 +95,29 @@ Firestore event          | DocumentEventData     | --trigger-event providers/clo
...
+`ICloudEventFunction<MessagePublishedData>`, the Functions Framework
+will adapt the incoming HTTP request to present it to your function
+as a CloudEvent, as if you had deployed via `--trigger-topic`. The
+requests for push subscriptions do not contain topic data, but if
+you create the push subcription with a URL of
+`https://<your-function>/projects/<project-id>/topics/<topic-id>`
+then the Functions Framework will infer the topic name from the path
+of the HTTP request. If the topic name cannot be inferred
+automatically, a topic name of
+`projects/unknown-project!/topics/unknown-topic!` will be used in
```
matthewrobertson, line 112: It seems like we aren't consistent about what we use here across languages.

- [In Node.js](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/test/integration/cloud_event.ts#L188) we just drop it and make the CE subject `//pubsub.googleapis.com/`
- [In Ruby](https://github.com/GoogleCloudPlatform/functions-framework-ruby/blob/main/test/test_legacy_event_converter.rb#L111) we set it to `//pubsub.googleapis.com/UNKNOWN_PUBSUB_TOPIC`

Also, I **think** it is might be possible to always extract the project name from the subscription in the payload.

matthewrobertson (approved):
LGTM!

docs/deployment.md
jskeet: > It seems like we aren't consistent about what we use here across languages.

Mmm. The reason I chose this is to keep structure that will work in the absence of "real" data, but make it obvious that it's *not* real data. That may or may not be a good choice. Let's discuss this more before actually releasing.

> Also, I think it is might be possible to always extract the project name from the subscription in the payload.

That assumes that the subscription is in the same project as the topic, which may not be the case. (I *think* cross-project PubSub subscriptions are a thing, with appropriate IAM. I could be wrong.) For the emulator that doesn't matter, but for the genuine PubSub push notifications, I think it's better to give either "obviously wrong" data or "user-specified data" (the URL path) rather than "looks plausible but may be wrong" data.

We should definitely talk about that part more. I'm going to merge this as-is, but file a new issue so that we don't forget about it before the next release.

PR #242, opened 2022-03-03
feat: Warn if the "unknown topic" resource name is used

### LINKED ISSUE #239: Decide how to handle unknown PubSub topics for push notifications
Currently we use a resource name of `projects/unknown-project!/topics/unknown-topic!` when we don't know the topic resource name from a PubSub push notification. (If the path on the incoming HTTP request looks like a topic name, we'll use that, so users can set it up in a useful way, but they may well not do so.)

This is inconsistent with other Functions Frameworks.

Options:

- Change to be consistent with at least one of the other languages before the next release
- Decide consciously to go with what we've got and release, and don't change it
- Stick with what we've got at the moment, but document that the format may change in future releases

jskeet:
Closes #239 (in conjunction with a discussion of Functions Frameworks owners)

Currently we warn on every request. Warning only once would be
feasible, but only globally, which isn't ideal.

### DIFF
--- a/src/Google.Cloud.Functions.Framework.Tests/GcfEvents/GcfConvertersTest.cs
+++ b/src/Google.Cloud.Functions.Framework.Tests/GcfEvents/GcfConvertersTest.cs
@@ -15,7 +15,10 @@
 using CloudNative.CloudEvents;
 using CloudNative.CloudEvents.SystemTextJson;
 using Google.Cloud.Functions.Framework.GcfEvents;
+using Google.Cloud.Functions.Testing;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using System;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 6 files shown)


jskeet:
Most of the change here is in tests - the change from using "providers/cloud.pubsub/eventTypes/topic.publish" to "google.pubsub.topic.publish" for the simulated type in the context is not a user-facing change - it just feels like we should use the more recent event type rather than the legacy one.

src/Google.Cloud.Functions.Framework.Tests/GcfEvents/GcfConvertersTest.cs
```diff
@@ -166,36 +165,36 @@ public async Task MinimalValidEvent_RawPubSub()
...
+        public async Task RawPubSubTopic_NoPath(string? path, string expectedTopicResourceName, bool expectWarning)
         {
+            var logger = new MemoryLogger("converter");
             var context = GcfEventResources.CreateHttpContext("raw_pubsub.json", path);
-            var cloudEvent = await GcfConverters.ConvertGcfEventToCloudEvent(context.Request, s_jsonFormatter);
+            var cloudEvent = await GcfConverters.ConvertGcfEventToCloudEvent(context.Request, s_jsonFormatter, logger);
             Assert.Equal($"//pubsub.googleapis.com/{expectedTopicResourceName}", cloudEvent.Source?.ToString());
             string expectedTopic = expectedTopicResourceName.Split('/').Last();
             Assert.Equal(expectedTopic, (string) cloudEvent["topic"]!);
+            Assert.Equal(expectWarning, logger.ListLogEntries().Any(entry => entry.Level >= LogLevel.Warning));
```
amanda-tarafa, line 183: I'm guessing "unknown topic" is the only thing we are warning on here, right?

amanda-tarafa (approved):
LGTM

src/Google.Cloud.Functions.Framework.Tests/GcfEvents/GcfConvertersTest.cs
jskeet: Yup.

PR #243, opened 2022-04-07
ci: test dotnet6

grant:
Testing the CI for dotnet6.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        dotnet: [ "3.1.x" ]
+        dotnet: [ "3.1.x", "6.0.x" ]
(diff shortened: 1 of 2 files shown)


jskeet:
This isn't actually testing with .NET 6. It's using the .NET 6 SDK to build, but the test target in the csproj hasn't failed. I'm happy to do that, but it's probably better for me to do it in a single PR.

grant:
> This isn't actually testing with .NET 6. It's using the .NET 6 SDK to build, but the test target in the csproj hasn't failed. I'm happy to do that, but it's probably better for me to do it in a single PR.

Thanks. Was doing some rudimentary testing.

I definitely do not currently have the knowledge to properly test this runtime version for this language.

Maybe this would be on the right track:

https://docs.microsoft.com/en-us/aspnet/core/migration/31-to-60?view=aspnetcore-6.0&tabs=visual-studio

PR #249, opened 2022-06-10
test: Creates a buildpack validation action

jskeet:
This will ensure that any changes to the Functions Framework are
compatible with the buildpack. In order to use the latest source
code, it builds the Functions Framework locally and includes the
package file in a temporary copy of the conformance tests, along
with a nuget.config file to use that local package.

### DIFF
--- a/.github/workflows/buildpack-integration-test.yaml
+++ b/.github/workflows/buildpack-integration-test.yaml
@@ -1,0 +1,27 @@
+# Validates Functions Framework with GCF buildpacks.
+name: Buildpack Integration Test
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  dotnet3:
+    uses: GoogleCloudPlatform/functions-framework-conformance/.github/workflows/buildpack-integration-test.yml@v1.5.2
+    with:
+      # This builds a local copy of the Functions Framework, and creates
+      # a standalone copy of the conformance test function in tmp
+      # with a NuGet config file to use the local FF.
+      prerun: 'src/Google.Cloud.Functions.ConformanceTests/create-standalone.sh'
+      http-builder-source: 'tmp/Google.Cloud.Functions.ConformanceTests'
+      http-builder-target: 'HttpFunction'
+      cloudevent-builder-source: 'tmp/Google.Cloud.Functions.ConformanceTests'
+      cloudevent-builder-target: 'UntypedCloudEventFunction'
+      builder-runtime: 'dotnet3'
+      # Latest uploaded tag from us.gcr.io/fn-img/buildpacks/dotnet3/builder
+      builder-tag: 'dotnet3_20220516_3_1_416_RC00'
+      # Ask the test runner to wait for 5 seconds for the server to start
+      # before sending it requests. It should come up much quicker
+      # than that, but 5 seconds will rule out "slow startup" as
+      # a failure mode.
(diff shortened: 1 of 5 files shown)


jskeet:
I'm going to temporarily make this run on PRs as well, just so we can test it...

jskeet:
@anniefu: It looks like the build has worked, but the conformance test itself has failed.

I wonder whether the conformance test runner is possibly not waiting long enough for the server to come up? If you look at the [conformance test script](https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/main/run-conformance-tests.sh) we wait for 5 seconds (as a "this should definitely be long enough" time) before firing the actual requests.

Given that the client itself has a `startDelay` parameter, can we expose that in the `buildpack-integration-test.yml` workflow?

anniefu:
Plumbed it through! Please, give it a try https://github.com/GoogleCloudPlatform/functions-framework-conformance/pull/106/files

jskeet:
Wahoo! Thanks very much, will try tomorrow :)

src/Google.Cloud.Functions.ConformanceTests/create-standalone.sh
```diff
@@ -0,0 +1,46 @@
...
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
+# This script is designed to be used in the buildpack validation
+# hook. Its purpose is to provide a function which can be built
+# using a buildpack, but with the latest (potentially unreleased)
+# version of the Functions Framework.
+
+# Steps:
+#
+# - Clear out tmp
```
anniefu, line 14: [nit] You could skip clearing temp and moving the function source code around if you like. You'll get a throwaway GitHub runner machine per job, so each test run will get its own copy of the repo and clean state.

jskeet: True, but it makes it harder to test - given that there's very little to copy, I'll keep it for now.

jskeet:
This is now working - I've removed a final commit which made it run the action on PRs as well, which I used just for testing.

.github/workflows/buildpack-integration-test.yaml
```diff
@@ -0,0 +1,27 @@
...
+      # with a NuGet config file to use the local FF.
+      prerun: 'src/Google.Cloud.Functions.ConformanceTests/create-standalone.sh'
+      http-builder-source: 'tmp/Google.Cloud.Functions.ConformanceTests'
+      http-builder-target: 'HttpFunction'
+      cloudevent-builder-source: 'tmp/Google.Cloud.Functions.ConformanceTests'
+      cloudevent-builder-target: 'UntypedCloudEventFunction'
+      builder-runtime: 'dotnet3'
+      # Latest uploaded tag from us.gcr.io/fn-img/buildpacks/dotnet3/builder
+      builder-tag: 'dotnet3_20220516_3_1_416_RC00'
+      # Ask the test runner to wait for 5 seconds for thserver to start
```
amanda-tarafa, line 23:
```suggestion
      # Ask the test runner to wait for 5 seconds for the server to start
```

amanda-tarafa (approved):
LGTM

.github/workflows/buildpack-integration-test.yaml
jskeet: Whoops, thanks.

PR #252, opened 2022-07-18
chore: Add 6.0.x to conformance and unit tests

MontyCarter:


jskeet (approved):
There's more we need to do for .NET 6, but I'm happy enough with this as a starting point.
(We may well end up removing the 3.1.x target fairly soon, but that can be a separate decision.)

jskeet:
I'm not quite sure why this has been merged now, but with no changes. Thinking about it further, we don't actually need a test matrix in CI to test against .NET 6.0. We can just make the normal test project test both. I'll do that later this week if I can.

PR #256, opened 2022-08-30
build: enable release-please

chingor13:
On-boards this repository onto release-please.

An example release pull request: https://github.com/chingor13/functions-framework-dotnet/pull/7

Towards https://github.com/googleapis/repo-automation-bots/issues/3280

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,0 +1,195 @@
+{
+  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+  "release-type": "simple",
+  "include-component-in-tag": true,
+  "include-v-in-tag": false,
+  "changelog-path": "docs/history.md",
+  "packages": {
+    "src/Google.Cloud.Functions.Framework": {
+      "component": "Google.Cloud.Functions.Framework",
+      "changelog-path": "/docs/history.md",
+      "extra-files": [
+        {
+          "type": "xml",
+          "path": "/src/CommonProperties.xml",
+          "xpath": "//Project/PropertyGroup/Version"
+        }
+      ]
+    },
+    "src/Google.Cloud.Functions.Hosting": {
+      "component": "Google.Cloud.Functions.Hosting",
+      "extra-files": [
+        {
+          "type": "xml",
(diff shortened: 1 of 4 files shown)


jskeet:
I notice the PR has reformatted all the XML project files, removing the space between the final attribute value and the `/>` in each element. How feasible would it be to not do that? While it's not too much of an issue for this specific repo, it could become much more of a problem in other repositories where the files are generated/edited by different tooling... or would we avoid release-please modifying those files, and instead just modify their source of truth?

release-please-config.json
```diff
@@ -0,0 +1,195 @@
...
+        {
+          "type": "xml",
+          "path": "/src/CommonProperties.xml",
+          "xpath": "//Project/PropertyGroup/Version"
+        }
+      ]
+    },
+    "src/Google.Cloud.Functions.Hosting": {
+      "component": "Google.Cloud.Functions.Hosting",
+      "extra-files": [
```
jskeet, line 21: Is there any way of globbing this? I can easily see us adding a new example but forgetting to update the version numbers. (I've already got a shell script to do the update, if release-please can run that as part of the process...) It's at https://github.com/GoogleCloudPlatform/functions-framework-dotnet/blob/main/update-project-references.sh

jskeet (commented):
I notice that the release PR has "Miscellaneous Chores" in some histories but not others - is there a way of excluding them from all histories?

release-please-config.json
chingor13: There's currently not a way to glob extra files, but I opened https://github.com/googleapis/release-please/issues/1619 to investigate

jskeet:
> We are using the @xmldom/xmldom library to parse/update the XML document, I'll investigate options on the trailing space, but I'm not optimistic.

Understood. 

Modifying the source of truth: yup. Happy to explore options there.

Miscellaneous chores: I hadn't spotted that that was all from a single synthetic commit. I guess if we have an extra script to fix up formatting etc, it could always remove those sections if that's the simplest way of getting to the right output :)

chingor13:
Closing in favor of #268

PR #257, opened 2022-09-01
docs: Add an example of deployment using a local NuGet package

jskeet:
Note that the inclusion of the .nupkg file here is deliberate, for
simplicity. We want the solution to open and work without any
pre-steps of building the .nupkg file and copying it to the right
place.


amanda-tarafa (approved):
LGTM

PR #264, opened 2022-10-14
chore: Install .NET Core in buildpack integration test

jskeet:
This will install the latest version, rather than using the default. We need this as we build the latest version of the Functions Framework for the integration test.


amanda-tarafa (approved):
LGTM

PR #266, opened 2022-10-14
fix: Revert .NET SDK version changes

jskeet:
We need to work out how to get the latest version of the SDK when running the buildpack integration test


jskeet:
Messed this up somehow...

PR #273, opened 2022-10-18
chore: Refactor release-please to be a single component

jskeet:
We release all of the Functions Framework together, even though it's four packages - like GAX. We don't want or need to keep track of the versions or history for the separate packages separately.


amanda-tarafa (approved):
LGTM

PR #277, opened 2022-10-24
feat: Prepare for .NET 6 / dotnet6 support

jskeet:
Please review (when you're back) one commit at a time.

The commit "chore: Change how a missing CloudEventAdapter is handled" needs the most attention, and I'd be very happy to discuss alternatives.

### DIFF
--- a/.github/workflows/buildpack-integration-test.yaml
+++ b/.github/workflows/buildpack-integration-test.yaml
@@ -6,7 +6,7 @@ on:
       - main
   workflow_dispatch:
 jobs:
-  dotnet3:
+  dotnet6:
     uses: GoogleCloudPlatform/functions-framework-conformance/.github/workflows/buildpack-integration-test.yml@v1.8.0
     with:
       # This builds a local copy of the Functions Framework, and creates
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -8,7 +8,7 @@ jobs:
   run-conformance:
     strategy:
       matrix:
-        dotnet: [ "3.1.x" ]
+        dotnet: [ "6.0.x" ]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
 
@@ -23,7 +23,7 @@ jobs:
       with:
         submodules: true
 
-    - name: Setup .NET Core
+    - name: Setup .NET
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: ${{ matrix.dotnet }}
--- a/src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
+++ b/src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
@@ -29,23 +29,26 @@ namespace Google.Cloud.Functions.Framework
         private readonly ILogger _logger;
 
         /// <summary>
+        /// Constructor without a <see cref="CloudEventFormatter"/>, whose sole purpose is to throw a clear exception
+        /// if an event formatter has not been configured.
+        /// </summary>
+        [Obsolete("This constructor only exists to enable a clear error message. It always throws an exception.", error: true)]
+        public CloudEventAdapter(ICloudEventFunction<TData> function, ILogger<CloudEventAdapter<TData>> logger) =>
+            throw new InvalidOperationException(
+                "No CloudEventFormatter has been configured for the CloudEvent function. " +
+                "The Functions Framework hosting will provide an event formatter if the target " +
+                "data type is decorated with CloudEventFormatterAttribute; otherwise, the formatter " +
+                "should be configured via dependency injection.");
+
+        /// <summary>
         /// Constructs a new instance based on the given CloudEvent Function.
         /// </summary>
         /// <param name="function">The CloudEvent Function to invoke.</param>
-        /// <param name="formatter">The CloudEvent formatter to use to deserialize the CloudEvent. This is expected to be a single-type formatter,
-        /// targeting <typeparamref name="TData"/>.</param>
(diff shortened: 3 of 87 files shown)


jskeet:
The "test (3.1.x, xyz-latest)" actions will never complete, as they no longer exist...

src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
```diff
@@ -31,22 +33,27 @@ public sealed class CloudEventAdapter<TData> : IHttpFunction where TData : class
...
+        /// to be thrown if no formatters are available. Only a single formatter is currently used, which will be the last
+        /// one in the sequence to allow for functions to add their own and override the default.
+        /// </remarks>
         /// <param name="function">The CloudEvent Function to invoke.</param>
-        /// <param name="formatter">The CloudEvent formatter to use to deserialize the CloudEvent. This is expected to be a single-type formatter,
-        /// targeting <typeparamref name="TData"/>.</param>
+        /// <param name="formatters">A sequence of CloudEvent formatters to use to deserialize the CloudEvent. See remarks for more details.</param>
         /// <param name="logger">The logger to use to report errors.</param>
-        public CloudEventAdapter(ICloudEventFunction<TData> function, CloudEventFormatter formatter, ILogger<CloudEventAdapter<TData>> logger)
+        public CloudEventAdapter(ICloudEventFunction<TData> function, IEnumerable<CloudEventFormatter> formatters, ILogger<CloudEventAdapter<TData>> logger)
```
amanda-tarafa, line 44: My initial reaction is that receiving a sequence for null-checking is just a fancy way of implementing a null-forgiving operator. In particular, because we really don't use a sequence at all, just the last element on it, if any. So, I'd probably go with just using the null-forgiving operator.

Still, two other alternatives to consider:

- I think, that if you create a second constructor that receieves no formatter, and only throws he exception, that will be called if there's no formatter registered. I'm pretty sure I've used this before. And then here you continue to receive a single formatter. Granted, having a constructor that only throws is very weird.
- Receive the single non-null formatter, and just let DI throw the exception that it normally throws when it can't resolve a service, and document that very well in a troubleshooting section of the docs.

amanda-tarafa (approved):
LGTMing as I'm fine if you'd like to merge as is. I did leave a couple of comments around the handling of the missing dependency.
Also, chech some of the config files for some possible reamining netcoreapp3.1 references?

.github/workflows/conformance.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       with:
         submodules: true
 
-    - name: Setup .NET Core
+    - name: Setup .NET
       uses: actions/setup-dotnet@v3
```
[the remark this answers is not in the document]
jskeet: That's the version of the GitHub action (https://github.com/actions/setup-dotnet) - that's fine :)

.github/workflows/buildpack-integration-test.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
       http-builder-target: 'HttpFunction'
       cloudevent-builder-source: 'tmp/Google.Cloud.Functions.ConformanceTests'
       cloudevent-builder-target: 'UntypedCloudEventFunction'
-      builder-runtime: 'dotnet3'
+      builder-runtime: 'dotnet6'
       # Latest uploaded tag from us.gcr.io/fn-img/buildpacks/dotnet3/builder
       builder-tag: 'dotnet3_20220516_3_1_416_RC00'
```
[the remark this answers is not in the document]
jskeet: Hmm... interestingly, there *isn't* a tag for dotnet6. Will ask @anniefu about that later.

src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
jskeet: I'd prefer not to go down the second option here - I definitely want us to be able to specify an error message.
My concern over using the null-forgiving operator is that DI might *start* validating that more aggressively later... we're effectively violating the contract and hoping they don't spot it.

I'll investigate the first option though - that sounds promising.

jskeet: I've added a new constructor, and it's working fine. The fourth commit contains that - PTAL.

src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
```diff
@@ -28,24 +28,27 @@ public sealed class CloudEventAdapter<TData> : IHttpFunction where TData : class
         private readonly CloudEventFormatter _formatter;
         private readonly ILogger _logger;
 
+        /// <summary>
+        /// Constructor without a <see cref="CloudEventFormatter"/>, whose sole purpose is to throw a clear exception
+        /// if an event formatter has not been configured.
+        /// </summary>
+        [Obsolete("This constructor only exists to enable a clear error message. It always throws an exception.", error: true)]
```
amanda-tarafa, line 35: Nice making it obsolete!

amanda-tarafa (approved):
LGTM!!

src/Google.Cloud.Functions.Framework/CloudEventAdapterTData.cs
jskeet: And "hard obsolete" at that - it's impossible to call other than via reflection.
We may want to add some more attributes to avoid it being shown by Intellisense - although I don't expect anyone to be creating this by hand anyway.

jskeet:
Will wait until we've seen the result of #278, and update this PR accordingly - hopefully I can remove the buildpack integration test change entirely.

amanda-tarafa (approved):
LGTM

PR #278, opened 2022-10-31
chore: Update buildpack integration test to use latest tag implicitly

jskeet:
Using the "dotnet" builder should work for both dotnet3 and dotnet6.


jskeet:
(Once we've merged this, we'll see whether it actually works...)

amanda-tarafa (approved):
LGTM

PR #296, opened 2023-03-28
feat: Implement strongly typed function signatures for dotnet

garethgeorge:
Implements a strongly typed function signature that provides automatic request deserialization and response serialization for dotnet.

### DIFF
--- a/src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
+++ b/src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
@@ -1,0 +1,31 @@
+// Copyright 2023, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.AspNetCore.Http;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Framework;
+
+/// <summary>
+/// Responsible for reading a strongly typed request object from an
+/// HTTP request.
+/// </summary>
+public interface IHttpRequestReader<TRequest>
+{
+    /// <summary>
+    /// Asynchronously reads a strongly typed request object.
+    /// </summary>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public Task<TRequest> ReadRequestAsync(HttpRequest httpRequest);
+}
--- a/src/Google.Cloud.Functions.Framework/ITypedFunction.cs
+++ b/src/Google.Cloud.Functions.Framework/ITypedFunction.cs
@@ -1,0 +1,36 @@
+// Copyright 2023, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Framework;
+
+/// <summary>
+/// A typed function accepting a structured request and asynchronously 
+/// returning a structured response.
+/// </summary>
+/// <typeparam name="TRequest">Type of the request object.</typeparam>
+/// <typeparam name="TResponse">Type of the response object.</typeparam>
+public interface ITypedFunction<TRequest, TResponse>
+{
+    /// <summary>
+    /// Asynchronously handles an incoming request, expected to return 
+    /// a structured response.
+    /// </summary>
+    /// <param name="request">The request payload, deserialized from the incoming request.</param>
+    /// <param name="cancellationToken">A cancellation token which indicates if the request is aborted.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    Task<TResponse> HandleAsync(TRequest request, CancellationToken cancellationToken);
+}
--- a/src/Google.Cloud.Functions.Framework/TypedFunctionAdapter.cs
+++ b/src/Google.Cloud.Functions.Framework/TypedFunctionAdapter.cs
@@ -1,0 +1,81 @@
+// Copyright 2023, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Functions.Framework;
+
+/// <summary>
+/// An adapter to implement an HTTP Function based on a <see cref="ITypedFunction{TRequest, TResponse}"/>,
+/// with built-in request deserialization and response serialization.
+/// </summary>
+public sealed class TypedFunctionAdapter<TRequest, TResult> : IHttpFunction
+{
+    private readonly ITypedFunction<TRequest, TResult> _function;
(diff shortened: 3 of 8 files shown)


jskeet:
Starting to review and fix nits locally now - but while I remember, we generally find it better to work on forks and create pull requests from there, rather than accumulating branches on the "main" fork here. Not a problem, but if you could create a fork to work on for the next PR, that'd be great :)

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -171,6 +172,19 @@ internal static IServiceCollection AddServicesForFunctionTarget(IServiceCollecti
...
+            else if (GetGenericInterfaceImplementationTypeArguments(functionType, typeof(ITypedFunction<,>)) is Type[] signatureTypes)
+            {
+                Type argType = signatureTypes[0];
+                Type retType = signatureTypes[1];
+
+                services
+                    .AddScoped(typeof(IHttpFunction), typeof(TypedFunctionAdapter<,>).MakeGenericType(signatureTypes))
+                    .AddScoped(typeof(ITypedFunction<,>).MakeGenericType(signatureTypes), functionType);
+
+                Type jsonFormatImpl = typeof(JsonFormatImpl<,>).MakeGenericType(argType, retType);
```
jskeet: I've added a comment here to indicate that users can specify their own readers/writers, and added a new test for that.

jskeet (commented):
I've added two extra commits with a bunch of changes - please review them carefully.
The first contains "real" changes (mostly naming, admittedly). The second is *just* changing to use file-scoped namespaces for the new code, but if I included that with the first commit, it would make it much harder to read.

If you're happy with the changes, we can then add another commit in the examples directory, then squash and merge.

src/Google.Cloud.Functions.Framework/TypedFunctionAdapter.cs
```diff
@@ -0,0 +1,85 @@
...
+            TRes res = await _function.HandleAsync(data, context.RequestAborted);
+
+            try
+            {
+                await _responseWriter.WriteResponseAsync(context.Response, res);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e.Message);
+                context.Response.StatusCode = StatusCodes.Status500InternalServerError;
```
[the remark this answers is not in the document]
garethgeorge: Right, I hadn't thought much about the fact that theres now no way to indicate a legitimate 500 error on the request parsing edge. For the initial implementation, it is my hope that *most* exceptions thrown in formatters will be legitimate parse errors or the such so we throw the 400 as a reasonable best guess.

@anniefu has also mentioned asks for exceptions that give some control over the HTTP status returned, it's something that would be neat to look at eventually.

src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.AspNetCore.Http;
+using System;
+
+namespace Google.Cloud.Functions.Framework
+{
+    /// <summary>
+    /// Responsible for reading a strongly typed request object from an
+    /// HTTP request.
+    /// </summary>
+    public interface IHttpRequestReader<TReq>
```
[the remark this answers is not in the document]
garethgeorge: It's an interesting thought, CancellationTokens are used to abort an operation if it's long running and the HTTP request dies? I'd say I'm also ambivalent. 

The cases I can imagine where it might be useful would be borderline applications of formatters i.e. augmenting the request with data fetched from a Cloud Storage bucket (or some similarly slow and expensive network operation you'd want to be able to cancel). In the typical case request decoding / response encoding should be so fast I can't imagine too much value.

WDYT?

garethgeorge:
Thanks Jon! Did a pass over your revisions, they LGTM. Appreciate the places you've revised the implementation and test coverage. 

Also good catch w.r.t. throwing an error on a null request payload in the default json reader, it's a good idea. That's something I'll aim to make sure we're consistent about (where we can be) across the other languages we're adding this feature to.

jskeet:
> Also good catch w.r.t. throwing an error on a null request payload in the default json reader, it's a good idea. 

Your code was already doing that, actually - I just added a test for it, and made the exception a bit clearer.

src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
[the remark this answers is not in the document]
jskeet: I agree that usually it'll be fast, in terms of processing. If a large result is generated, then it would be good to abort that early, but I'd expect that if the request is aborted, that output stream will become invalid anyway.

Let's go without for now... introducing it later would be a breaking change, but we can always introduce other interfaces to implement, and work things out that way.

jskeet (commented):
Given that a fair amount of the code is now mine, I'll ask for a second opinion to actually approve before we squash and merge.

src/Google.Cloud.Functions.Framework/ITypedFunction.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020, Google LLC
```
amanda-tarafa, line 1: nit: 2023, and maybe check elsewhere

src/Google.Cloud.Functions.Framework/ITypedFunction.cs
```diff
@@ -0,0 +1,38 @@
...
+    public interface ITypedFunction<TReq, TRes> 
+      where TReq : class
+      where TRes : class
+    {
+        /// <summary>
+        /// Asynchronously handles an incoming request, expected to return 
+        /// a structured response.
+        /// </summary>
+        /// <param name="data">Request payload</param>
+        /// <returns></returns>
```
amanda-tarafa: nit: empty docs

src/Google.Cloud.Functions.Framework/ITypedFunction.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// A typed function accepting a structured request and asynchronously 
+    /// returning a structured response.
+    /// </summary>
+    /// <typeparam name="TReq">type of the request object</typeparam>
+    /// <typeparam name="TRes">type of the response object</typeparam>
+    public interface ITypedFunction<TReq, TRes> 
+      where TReq : class
+      where TRes : class
```
amanda-tarafa: Do we really need these constraints?

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
```diff
@@ -171,6 +173,22 @@ internal static IServiceCollection AddServicesForFunctionTarget(IServiceCollecti
...
+            else if (GetGenericInterfaceImplementationTypeArguments(functionType, typeof(ITypedFunction<,>)) is Type[] signatureTypes)
+            {
+                Type requestType = signatureTypes[0];
+                Type responseType = signatureTypes[1];
+
+                services
+                    .AddScoped(typeof(IHttpFunction), typeof(TypedFunctionAdapter<,>).MakeGenericType(signatureTypes))
+                    .AddScoped(typeof(ITypedFunction<,>).MakeGenericType(signatureTypes), functionType);
+
+                // Provide JSON reader/writer implementations using System.Text.Json by default.
```
amanda-tarafa, line 185: thought: This seems a little random? Maybe consider adding a utility for generating the System.Text.Json reader/writer and find a way to annotate the typed function to specify those. It would be similar to the formatter for Cloud Events.

src/Google.Cloud.Functions.Framework/ITypedFunction.cs
amanda-tarafa: Oh never mind, I've seen these have been removed in later commits.

src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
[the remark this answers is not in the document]
amanda-tarafa: There's no need to explicitly pass the cancellation token. HttpRequest and HttpResponse both have access to the HttpContext, which in turn has the `CancellationToken HttpContext.RequestAborted` property.

The alternative I would consider for the reader and the writer is to actually pass the HttpContext directly. That makes it clearer that the whole thing is available for using in serialization/deserialization. There are other elements of HttpContext that may be useful for deserialization, like the Items collection.

src/Google.Cloud.Functions.Framework/ITypedFunction.cs
jskeet: Thanks, will fix - it looks like we're missing headers in another couple of files too.

src/Google.Cloud.Functions.Hosting/HostingInternals.cs
jskeet: Unlike the CloudEvents scenario, we don't have any common non-FF assembly that could contain an attribute specifying what reader/writer should be used.

But we're expecting JSON to be the default serialization format for most things. We may provide docs with an example of Json.NET, as that'll be reasonably common too.

src/Google.Cloud.Functions.Framework/IHttpRequestReader.cs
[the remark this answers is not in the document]
jskeet: Ah, hooray. Good to know we don't need to pass anything else.

But I prefer keeping the request/response directly - it emphasizes what the code is expected to work with. (It would be relatively unusual for a request reader to need to reach into the context, for example.) So long as they *can* do it, I think that's fine.

jskeet (approved):
All copyright files checked now. I've squashed it into a single commit - will rebase and merge shortly. Expect to see buildpack integration tests failing - that's due to our SDK being a bit later than the one in buildpacks at the moment.

PR #298, opened 2023-04-12
chore: Create renovate bot config

kenneth-rosario:


.github/renovate.json
```diff
@@ -0,0 +1,15 @@
...
+      "description": "Create a PR whenever there is a new major version",
+      "matchUpdateTypes": [
+        "major"
+      ]
+    }
+  ],
+  "ignorePaths": [
+    "examples/**"
+  ]
+}
```
jskeet, line 15:
```suggestion
{
  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
  "extends": [
    "config:base",
    ":disableDependencyDashboard"
  ],
  "ignorePaths": [
    ".kokoro/**"
  ],
  "schedule": ["before 8am"],
  "timezone": "Europe/London"
}
```

That's close to what we're using in https://github.com/googleapis/google-cloudevents-dotnet - but in .NET we've been tending to only update dependencies where we really need to. The suggested configuration is likely to suggest more changes than we actually want, so will need tweaking over time, but is a good starting point.

jskeet (commented):
Not sure why we didn't do FF when we configured the other repos...

jskeet (approved):

PR #304, opened 2023-04-14
chore: Create blunderbuss.yml according to preference

kenneth-rosario:


jskeet (approved):
I'm okay with this for now, but I'm still expecting to personally review any changes at least for a while

PR #323, opened 2023-04-20
tests: Fix buildpack integration tests

jskeet:
We need to build the code with whatever .NET SDK is installed in the buildpack. We shouldn't force that to be the same SDK that the Functions Framework itself uses for its normal build.


jskeet:
The test isn't run as part of a PR, so I've run it manually against an equivalent branch I pushed to this main fork: https://github.com/GoogleCloudPlatform/functions-framework-dotnet/actions/runs/4751597479

(The warnings are from actions in the FF conformance repo itself.)

amanda-tarafa (approved):
LGTM

PR #381, opened 2023-08-04
chore: Update .NET SDK

jskeet:
(We'll need to see if the conformance tests pass...)


jskeet:
Whoops, we already had a renovate PR.

PR #499, opened 2024-04-16
docs: Deployment instructions for .NET 8

### LINKED ISSUE #444: Document deployment in .NET 8
There's a `dotnet8` runtime, currently in beta. I've tested that it works with a very simple function:

- Use the normal template (gcf-http)
- Edit the project file to change `net6.0` to `net8.0` (for `TargetFramework`)
- Use functionality only available in .NET 8 (some JSON options in this particular case)
- Deploy with the command below

```sh
$ gcloud beta functions deploy test-dotnet-8 --runtime=dotnet8 --entry-point=TestDotnet8.Function \
    --trigger-http --allow-unauthenticated --gen2
```

It worked first time...

Note that it's gen2 only.

It's possible (likely?) that this is still using the .NET 6 version of ASP.NET Core, as that's what the Functions Framework targets - but adding an explicit dependency *might* address that; more work to be done there.

At some point we should document all of this, and maybe eventually change the templates.

jskeet:
Note that we don't have any examples targeting .NET 8, partly as we're still building on .NET 6.

Fixes #444.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -63,7 +63,7 @@ triggered.
 HTTP functions are deployed using `--trigger-http`. For example:
 
 ```text
-gcloud functions deploy hello-functions --runtime dotnet3 --trigger-http --entry-point HelloFunctions.Function
+gcloud functions deploy hello-functions --runtime dotnet6 --trigger-http --entry-point HelloFunctions.Function
 ```
 
 On successful deployment, details of the deployed function will be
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -63,7 +63,7 @@ triggered.
 HTTP functions are deployed using `--trigger-http`. For example:
 
 ```text
-gcloud functions deploy hello-functions --runtime dotnet3 --trigger-http --entry-point HelloFunctions.Function
+gcloud functions deploy hello-functions --runtime dotnet6 --trigger-http --entry-point HelloFunctions.Function
```
jskeet, line 66: This line is just fixing a legacy issue.

amanda-tarafa (approved):

PR #540, opened 2024-06-03
chore: Observe the cancellation token in samples and templates

jskeet:


jskeet:
(Updated a lot more samples, to make sure that context.Response.WriteAsync always has the cancellation token...)

amanda-tarafa (approved):

